PR #8, opened 2020-01-22
EG-226: Added model to represent Ocelot meta details

cmrhmrc:


test/base/ProcessJson.scala
```diff
@@ -0,0 +1,1941 @@
...
+
+  val invalidJson: JsValue = Json.parse(
+    """
+      |{
+      | "invalidJson": "invalid Json"
+      |}
+    """.stripMargin
+  )
+
+  val fullJson: JsValue = Json.parse(
```
macgrewal: The JSON is fine for testing actual processes but to test parsing logic is unessasaily large. Possibly have multiple valid JSON payloads named after the purpose would be easier for new people to understand/maintain?

PR #9, opened 2020-01-22
EG-229 Define CalloutStanza class and implement Unit tests

woollett-r:
Define CalloutStanza under sealed trait Stanza and create unit tests.


app/models/ocelot/Stanza.scala
```diff
@@ -0,0 +1,48 @@
...
+import play.api.libs.json._
+import play.api.libs.functional.syntax._
+
+sealed trait Stanza
+
+
+/**
+ * Define Callout Stanza
+ */
+case class CalloutStanza(
```
macgrewal: Number of Stanzas is likely to grow upto 20 and this would make the file rather bloated. Please move stanza models into the own files

app/models/ocelot/Stanza.scala
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.ocelot
+
+import play.api.libs.json._
+import play.api.libs.functional.syntax._
+
+sealed trait Stanza
```
macgrewal: No need to make this model sealed

app/models/ocelot/Stanza.scala
```diff
@@ -0,0 +1,48 @@
...
+
+
+/**
+ * Define Callout Stanza
+ */
+case class CalloutStanza(
+                        noteType: String,
```
macgrewal: consider making `noteType` a type (e.g. `sealed trait CalloutType`) allowing matching on the callout types more specific than a text match.

test/models/ocelot/CalloutStanzaSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+ */
+
+package models.ocelot
+
+import org.scalatest._
+import play.api.libs.json._
+
+import scala.util.{Failure, Success, Try}
+
+class CalloutStanzaSpec extends FlatSpec {
```
macgrewal: `WordSpec` is the normal style of writing tests on the platform

test/models/ocelot/CalloutStanzaSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+  "read valid Callout stanza of type subtitle" should "Create a Callout stanza with a note type of Subtitle" in {
+
+    val subTitleCalloutStanzaJson : JsValue = Json.parse( subTitleCalloutStanzaInputJson )
+
+    val subTitleCalloutStanza: CalloutStanza = subTitleCalloutStanzaJson.as[CalloutStanza]
+
+    assert( subTitleCalloutStanza == CalloutStanza( "Subtitle", 1, Seq( "2"), true ) )
+  }
+
+  "read Callout stanza with missing property" should "Cause an exception to be raised" in {
```
macgrewal: Story is about parsing valid JSON so this test was technically unnecessary. Though I happy to let it slide 😜

PR #21, opened 2020-02-04
Eg 220, Fix cyclomatic complexity > 10

cmrhmrc:


app/services/FlowErrors.scala
```diff
@@ -20,8 +20,8 @@ import models.ocelot.stanzas.Stanza
 
 trait FlowError
 
-case class UnknownStanza(unknown: Stanza) extends FlowError
-case class NoSuchPage(id: String) extends FlowError
+case class UnknownStanzaType(unknown: Stanza) extends FlowError
+case class StanzaNotFound(id: String) extends FlowError
 case class EmptyPage(id: String) extends FlowError
```
woollett-r: This error is not currently being used.

PR #31, opened 2020-02-14
Eg-278, Page, Paragraph, Link, TextItems twirl templates, removal of compiler warnings

cmrhmrc:


app/views/components/link.scala.html
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import models.ui.HyperLink
+
+@(link: HyperLink)(implicit messages: Messages)
+
+<a class="govuk-link" href="@link.dest" @{if (link.window) "target=_blank"}>@link.txt.value(messages.lang)</a>
```
woollett-r: Should the target definition be target="_blank" rather than "target=_blank"

app/views/components/page.scala.html
```diff
@@ -0,0 +1,36 @@
...
+
+@for( cmp <- pge.components) {
+  @{
+    cmp match {
+      case p: Paragraph => components.paragraph(p)
+      case l: HyperLink => components.link(l)
+      case h: H1 => components.h1_heading(h)
+      case h: H2 => components.h2_heading(h)
+      case h: H3 => components.h3_heading(h)
+      case x => {x}
```
woollett-r: I am not certain what the default action really should be. This is something we should discuss at a later date.

app/views/render_page.scala.html
```diff
@@ -24,17 +23,7 @@
 
 @layout(messages("dummy.title")) {
```
woollett-r: The message key "dummy.title" is hardcoded into the render. If we are going to use this template going forward I think the key should be passed in by the controller.

app/views/components/link.scala.html
cmrhmrc: I started with target="_blank"  and found it produced multiple quotes, the "target=_blank" syntax with the if statement produced the correct result

app/views/render_page.scala.html
cmrhmrc: Mac mentioned there is some other way of doing this, so I will remove the line and push to the branch

test/views/components/PageSpec.scala
```diff
@@ -0,0 +1,89 @@
...
+import org.scalatest.{Matchers, WordSpec}
+import org.scalatestplus.play.guice.GuiceOneAppPerSuite
+import play.api.inject.Injector
+import play.api.i18n.{Messages, MessagesApi, Lang}
+import play.api.test.FakeRequest
+import play.twirl.api.Html
+import org.jsoup.Jsoup
+import views.html.components.page
+import models.ui.{Page,H1,Paragraph,Text}
+import org.jsoup.nodes.{Document, Element}
```
woollett-r: IntelliJ is reporting Element is unused.

cmrhmrc: OK all done

test/views/components/PageSpec.scala
```diff
@@ -0,0 +1,89 @@
...
+      paras.first.text shouldBe openingPara.welsh
+
+      val h1s = doc.getElementsByTag("h1")
+      h1s.size shouldBe 1
+      h1s.first.text shouldBe title.welsh
+    }
+
+  }
+
+}
```
woollett-r: To fully test the page renderer we should really have a hyperlink on the page.

PR #47, opened 2020-03-03
EG-scalafmt, reformatted

cmrhmrc:


woollett-r:
Still appears to work.

PR #56, opened 2020-03-30
Eg-336, Add basic Error handling (no answer selected) to Question form

cmrhmrc:


app/models/ui/FormData.scala
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.ui
+
+import play.api.data.{Form, FormError}
```
woollett-r: Can you review the imports. I don't think they are all necessary.

app/models/ui/ErrorCallout.scala
```diff
@@ -0,0 +1,19 @@
+/*
```
woollett-r: I am not sure about the names of these two classes. "ErrorCallout" shows its origin in Ocelot. I be tempted to use ErrorMsg for ErrortCallout because it is just a container for an error message and then use something like KeyedErrorMsg for the class linking the error message to an element on the Html page.

app/models/ui/FormData.scala
cmrhmrc: will do

app/models/ui/ErrorCallout.scala
cmrhmrc: ErrorMsg is the real UI element going through to the template and contains an id from UI derived FormError. The ErrorCallout is a placeholder for a Callout of type Error which may or may not be turned into an ErrorMsg. The idea being when Questions are built, they will always contain the ErrorCallout objects, however ErroMsgs are only created from the ErrorCallouts if relevant errors are found in the FormData.

app/views/components/question.scala.html
```diff
@@ -24,6 +24,8 @@
     <h1 class="govuk-fieldset__heading">@components.render_text(q.text)</h1>
   </legend>
 
+  @components.error_message(q.errorMsgs)
```
woollett-r: I think this may be in the wrong place. I think the individual error messages should appear just above the  Html element associated with the errors. Therefore, the error messages should appear between the textual UI components and the radio buttons. See https://design-system.service.gov.uk/components/error-message/.

cmrhmrc: Yes will look at this (re. hangouts discussion)

test/views/components/ErrorSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+
+      val lists = doc.getElementsByTag("ul").asScala.toList
+      val listItems = lists(0).getElementsByTag("li").asScala.toList
+
+      listItems(0).text() shouldBe errorStrings(0)
+    }
+
+  }
+
+  "error_message" must {
```
woollett-r: You could add another test here for the visually hidden text. You would probably have to identify the element by its class.

PR #58, opened 2020-04-02
EG-354, restructured question rendering and added bullet points as hints

cmrhmrc:


test/views/components/QuestionSpec.scala
```diff
@@ -51,9 +51,15 @@ class QuestionSpec extends WordSpec with Matchers with GuiceOneAppPerSuite {
...
+    val leading = Text("You can buy", "Gwallwch brynu")
+    val bp1 = Text("apples", "afalau")
+    val bp2 = Text("oranges", "orennau")
+    val bp3 = Text("pears", "gellyg")
+    val bpList: BulletPointList = BulletPointList(leading, Seq(bp1, bp2, bp3))
+
     val answers = Seq(a1, a2, a3)
     val horizontalAnswers = Seq(a1.copy(hint = None), a2.copy(hint = None))
-    val question = Question(Text(q1), Seq(para1), answers)
+    val question = Question(Text(q1), Seq(bpList,para1), answers)
```
woollett-r, line 62: As you have added a bullet point list to the components in the question it would be good to check the bullet point list is in the html document.

app/views/components/bullet_point_list.scala.html
```diff
@@ -14,14 +14,18 @@
  * limitations under the License.
  *@
 
-@import models.ui.BulletPointList
+@import models.ui.{Paragraph, BulletPointList}
 
-@(bulletPointList: BulletPointList)( implicit messages: Messages )
+@(bulletPointList: BulletPointList, asHint: Boolean = false)( implicit messages: Messages )
```
woollett-r, line 19: Do we need a test for the hint option?

app/views/components/paragraph.scala.html
```diff
@@ -16,9 +16,11 @@
 
 @import models.ui.Paragraph
 
-@(para: Paragraph)(implicit messages: Messages)
+@(para: Paragraph, asHint: Boolean = false)(implicit messages: Messages)
```
woollett-r, line 19: Do we need a test for the hint option?

app/views/components/question_hint.scala.html
```diff
@@ -17,6 +17,17 @@
 @import models.ui._
 @import controllers.routes._
 
-@(p: Paragraph, id: String)(implicit request: Request[_], messages: Messages)
+@(questionBody: Seq[UIComponent], name: String)(implicit request: Request[_], messages: Messages)
```
woollett-r, line 20: Do we need an explicit test for this component?

app/views/components/bullet_point_list.scala.html
woollett-r: A test has been added.

app/views/components/paragraph.scala.html
woollett-r: Tests have been added

test/views/components/QuestionSpec.scala
woollett-r: Fixed

app/views/components/question_hint.scala.html
woollett-r: Leave with cover from tests for question component.

PR #60, opened 2020-04-03
Eg 355 - Fix bullet point error when non-whitespace character immediately follows placeholder annotation

woollett-r:
The fix only required changes to BulletPointBuilder and BulletPointListSpec. The other changes were the result of running ScalafmtAll on the code base.


app/services/BulletPointBuilder.scala
```diff
@@ -54,42 +54,34 @@ object BulletPointBuilder {
...
+  def fragmentsToDisplayAsSeq(str: String): Seq[String] = {
     val isEmpty: String => Boolean = _.isEmpty
 
     val (txts, matches) = TextBuilder.placeholderTxtsAndMatches(str)
     val matchTexts = matches.map(m => Option(m.group(1)).getOrElse(m.group(2)))
-    TextBuilder.merge(txts, matchTexts, Nil, isEmpty).mkString.split(' ').toList
+
+    val mergedTexts: Seq[String] = TextBuilder.merge(txts, matchTexts, Nil, isEmpty)
+
+    mergedTexts.filter(_ != "")
```
woollett-r, line 65: In response to your comment Chris I have modified the method to return a sequence of strings and changed the client methods appropriately. I haven't compressed the code quite as much as you requested as I find it useful to examine the merged texts with the debugger.

PR #61, opened 2020-04-06
Eg-354, reworked question page structure

cmrhmrc:


test/views/components/QuestionSpec.scala
```diff
@@ -69,50 +69,6 @@ class QuestionSpec extends WordSpec with Matchers with GuiceOneAppPerSuite {
     implicit override def messages: Messages = messagesApi.preferred(Seq(Lang("cy")))
```
woollett-r, line 69: Now that the generated HTML varies depending on whether, or not, the body is populated we could really do with a test for a question with an empty body.

PR #64, opened 2020-04-07
Eg-362, add connector to published process

cmrhmrc:


app/connectors/GuidanceConnector.scala
```diff
@@ -34,13 +34,19 @@ class GuidanceConnector @Inject() (httpClient: HttpClient, appConfig: AppConfig)
   def getProcess(id: String)(implicit hc: HeaderCarrier, context: ExecutionContext): Future[Option[Process]] =
```
woollett-r, line 34: How long should we keep this?

PR #65, opened 2020-04-09
EG-390, added route, controller, service and tests for publish

cmrhmrc:


app/services/GuidanceService.scala
```diff
@@ -52,6 +52,9 @@ class GuidanceService @Inject() (connector: GuidanceConnector, sessionRepository
   def scratchProcess(uuid: String, repositoryId: String)(implicit hc: HeaderCarrier, context: ExecutionContext): Future[Option[String]] =
     startProcessView(uuid, repositoryId, connector.scratchProcess)
 
+  def publishedProcess(processId: String, repositoryId: String)(implicit hc: HeaderCarrier, context: ExecutionContext): Future[Option[String]] =
+    startProcessView(processId, repositoryId, connector.publishedProcess)
```
woollett-r, line 56: The variable repositoryId is really a reference for the sessionId. Perhaps it would be better to use sessionId. When I see repositoryId I think of an identifier for a repository.

cmrhmrc: It is the id that will be used against the process in the mongo session repository

woollett-r: The variable has a dual use. Originally it is the session id, but at the service level it can be thought of as a key to the local data. Closing issue.

PR #68, opened 2020-04-16
Eg 355

woollett-r:
EG-355 Fix error when non-whitespace character immediately follows placeholder annotation

### DIFF
--- a/app/services/BulletPointBuilder.scala
+++ b/app/services/BulletPointBuilder.scala
@@ -101,60 +101,61 @@ object BulletPointBuilder {
     * @return The method returns the text and match components that contain the matched words
     */
   def locateTextsAndMatchesContainingLeadingText(
-                                                  noOfWordsToMatch: Int,
-                                                  inputTexts: List[String],
-                                                  textsProcessed: Int,
-                                                  inputMatches: List[Match],
-                                                  matchesProcessed: Int,
-                                                  outputTexts: List[String],
-                                                  outputMatches: List[Match],
-                                                  wordsProcessed: Int
-                                                ): (Int, List[String], List[Match]) = {
+      noOfWordsToMatch: Int,
+      inputTexts: List[String],
+      textsProcessed: Int,
+      inputMatches: List[Match],
+      matchesProcessed: Int,
+      outputTexts: List[String],
+      outputMatches: List[Match],
+      wordsProcessed: Int
+  ): (Int, List[String], List[Match]) = {
 
-      if (textsProcessed == inputTexts.size) {
-        if (matchesProcessed == inputMatches.size) {
-          (wordsProcessed, outputTexts, outputMatches)
-        } else {
-          locateMatchesContainingLeadingText(
-            noOfWordsToMatch,
-            inputTexts,
-            textsProcessed,
-            inputMatches,
-            matchesProcessed,
-            outputTexts,
-            outputMatches,
-            wordsProcessed
-          )
-        }
+    if (textsProcessed == inputTexts.size) {
+      if (matchesProcessed == inputMatches.size) {
+        (wordsProcessed, outputTexts, outputMatches)
       } else {
-        val text: String = inputTexts(textsProcessed)
-        val noOfWords: Int = wordsInText(text, textsProcessed, inputMatches, matchesProcessed)
-        if (processNextMatch(noOfWordsToMatch, wordsProcessed, noOfWords, inputTexts, textsProcessed, inputMatches, matchesProcessed)) {
-          locateMatchesContainingLeadingText(
-            noOfWordsToMatch,
-            inputTexts,
-            textsProcessed + 1,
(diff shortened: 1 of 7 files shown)


cmrhmrc:
Looking at UIBuilder.fromInstructionGroup. You determine the first and second bps, then do the rest with a call to createRemainingItems. Couldnt this call do all of the Bps?

app/services/BulletPointBuilder.scala
```diff
@@ -102,28 +106,42 @@ object BulletPointBuilder {
   def locateTextsAndMatchesContainingLeadingText(
```
cmrhmrc, line 106: This fails the > 50 lines test. Its a difficult fn to follow, the boolean switch argument and the nested ifs compound situation. Splitting into two functions to remove boolean switch arg would produce two simpler fns

app/services/BulletPointBuilder.scala
```diff
@@ -102,28 +106,42 @@ object BulletPointBuilder {
...
       noOfWordsToMatch: Int,
       inputTexts: List[String],
+      textsProcessed: Int,
       inputMatches: List[Match],
+      matchesProcessed: Int,
       outputTexts: Seq[String],
```
cmrhmrc, line 112: There's a mixing of Lists and Seqs and subsequent toLists, would be better to stick with one or other and remove need for conversions

app/services/BulletPointBuilder.scala
```diff
@@ -172,31 +192,42 @@ object BulletPointBuilder {
...
       texts: List[String],
+      textsProcessed: Int,
       matches: List[Match],
+      matchesProcessed: Int,
       items: List[String],
       matchText: Boolean,
       wordsProcessed: Int
   ): List[String] = {
 
     if (matchText) {
```
cmrhmrc, line 203: like locateTextsAndMatchesContainingLeadingText() , this uses a boolean switch arg, to make the fn do two different things; work on texts or work on matches, would be clearer to create two fns

app/services/BulletPointBuilder.scala
```diff
@@ -210,14 +241,35 @@ object BulletPointBuilder {
     *
     * @return - Returns a sub-sample of the input text
     */
-  def extractLeadingMatchedWords(noOfMatchedWords: Int, text: String): String = {
+  def extractLeadingMatchedWords(noOfMatchedWords: Int, texts: List[String], textsProcessed: Int, matches: List[Match], matchesProcessed: Int): String = {
+
+    val textElements: List[Match] = notSpaceRegex.findAllMatchIn(texts(textsProcessed)).toList
```
cmrhmrc, line 246: think this is the same without the need to draw on regex and Matches
texts(textsProcessed).split(" +").toList

app/services/BulletPointBuilder.scala
```diff
@@ -210,14 +241,35 @@ object BulletPointBuilder {
...
 
-    val matches: List[Match] = notSpaceRegex.findAllMatchIn(text).toList
+    if (matches.size == 0) {
```
cmrhmrc, line 248: more usual to match a list with a case Nil =>

app/services/BulletPointBuilder.scala
woollett-r: Done

app/services/BulletPointBuilder.scala
woollett-r: Split function into two although this did mean relinquishing the tailrec annotation.

app/services/BulletPointBuilder.scala
woollett-r: I've split the function into two as requested.

app/services/BulletPointBuilder.scala
woollett-r: Running the test class the outputs of notSpaceRegex.findAllMatchIn(texts(textsProcessed)).toList and texts(textsProcessed).split(" +").toList are not always equal. In addition using a regex simplifies the extraction of the final element of the leading text. As such I recommend with keep the existing code.

app/services/BulletPointBuilder.scala
```diff
@@ -247,4 +299,96 @@ object BulletPointBuilder {
   def getMatchText(m: Match): String = Option(m.group(1)).getOrElse(m.group(2))
 
   def wordsInString(text: String): Int = notSpaceRegex.findAllMatchIn(text).toList.size
```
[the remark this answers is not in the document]
woollett-r: As previously, the two approaches do not always give the same result so changing to the text splitting approach would require a rework.

app/services/BulletPointBuilder.scala
woollett-r: I've rewritten the method use Scala match/case statements. I note, however, that I had to break the method into two to reduce the cyclomatic(?) complexity.

PR #69, opened 2020-04-21
EG-398, replace Option based returns with Either in Service and Conector

cmrhmrc:


app/controllers/GuidanceController.scala
```diff
@@ -41,33 +43,47 @@ class GuidanceController @Inject() (
...
         pageContext.page match {
           case page: StandardPage => Ok(standardView(page, pageContext.processStartUrl))
           case page: QuestionPage => Ok(questionView(page, pageContext.processStartUrl, questionName(path), formProvider(questionName(path))))
         }
-      case None =>
-        Logger.warn(s"Request for PageContext at /$path returned nothing resulting in BadRequest")
+      case Left(NotFoundError) =>
+        logger.warn(s"Request for PageContext at /$path returned NotFound, returning NotFound")
+        NotFound(errorHandler.notFoundTemplate)
+      case badRequesterr @ Left(BadRequestError | InvalidProcessError) =>
```
woollett-r, line 59: Looking at the service method getPageContext I think it can raise InvalidProcessError and NotFound. Further, the repository can raise DatabaseError. As such I don't think we need to test for BadRequest. An InvalidProcessError would suggest something is wrong with the process returned from Mongo so perhaps we should return InternalServerError.

app/controllers/GuidanceController.scala
```diff
@@ -41,33 +43,47 @@ class GuidanceController @Inject() (
...
+        logger.warn(s"Request for PageContext at /$path returned ${badRequesterr.toString} returning BadRequest")
         BadRequest(errorHandler.notFoundTemplate)
+      case Left(err) =>
+        logger.error(s"Request for PageContext at /$path returned $err, returning InternalServerError")
+        InternalServerError(errorHandler.notFoundTemplate)
```
woollett-r, line 64: I'd use internalServerErrorTemplate.

app/repositories/SessionRepository.scala
```diff
@@ -67,22 +70,28 @@ class DefaultSessionRepository @Inject() (config: AppConfig, component: Reactive
...
   }
 
-  def set(key: String, process: Process): Future[Option[Unit]] =
-    insert(DefaultSessionRepository.SessionProcess(key, process.meta.id, process, DateTime.now(DateTimeZone.UTC))) map (_ => Some(())) recover {
-      case lastError =>
-        logger.error(s"Unable to persist process=${process.meta.id} against id=$key")
-        None
-    }
+  def set(key: String, process: Process): Future[RequestOutcome[Unit]] = {
+    val sessionDocument = Json.toJson(DefaultSessionRepository.SessionProcess(key, process.meta.id, process, DateTime.now(DateTimeZone.UTC)))
```
woollett-r, line 85: Can you use the type SessionProcess here?

app/repositories/SessionRepository.scala
```diff
@@ -67,22 +70,28 @@ class DefaultSessionRepository @Inject() (config: AppConfig, component: Reactive
...
 
+    collection
```
woollett-r, line 87: Following your recent observations should we be using "collection" here? Is there an appropriate method in ReactiveRepository?

app/services/GuidanceService.scala
```diff
@@ -20,54 +20,73 @@ import connectors.GuidanceConnector
...
               logger.warn(s"PageBuilder error $err for url $url on process ${process.meta.id}")
-              None
+              Left(InvalidProcessError)
             },
             pages => {
               val stanzaIdToUrlMap: Map[String, String] = pages.map(page => (page.id, s"/guidance${page.url}")).toMap
               pages
                 .find(page => page.url == url)
-                .map(pge => PageContext(uiBuilder.fromStanzaPage(pge, formData)(stanzaIdToUrlMap), s"/guidance${pages.head.url}"))
+                .fold(Left(NotFoundError): RequestOutcome[PageContext]) { pge =>
```
woollett-r, line 50: I'm not quite certain what to use here. The url comes from the redirect so I wouldn't expect this to fail. Maybe use internal server error as this should not happen.

app/services/GuidanceService.scala
```diff
@@ -20,54 +20,73 @@ import connectors.GuidanceConnector
...
+                  Right(PageContext(uiBuilder.fromStanzaPage(pge, formData)(stanzaIdToUrlMap), s"/guidance${pages.head.url}"))
+                }
             }
           )
-      }
+      case Left(err) =>
+        logger.warn(s"Repository returned $err, when attempting retrive process using id $sessionId")
```
woollett-r, line 56: Retrive should be retrieve.

app/controllers/GuidanceController.scala
cmrhmrc: Being called without a sessionId will pass through this route., this seems to be a valid BadRequestError

app/controllers/GuidanceController.scala
cmrhmrc: Yes, well spotted

app/repositories/SessionRepository.scala
cmrhmrc: yes?

app/repositories/SessionRepository.scala
cmrhmrc: there doesnt seem to anything suitable on the main class and this one doesnt generate the deprecated warning

app/services/GuidanceService.scala
cmrhmrc: The Url could have come from someone just trying random urls after a session has started

app/services/GuidanceService.scala
cmrhmrc: Yes!

app/repositories/SessionRepository.scala
woollett-r: Ok

app/repositories/SessionRepository.scala
woollett-r: Ok, I would say if you've gone to the trouble of creating a special type you should use it wherever necessary.

woollett-r: Type declaration removed.

PR #71, opened 2020-04-27
EG-365: Updated play-frontend-* dependencies and added timeout dialog

macgrewal:


conf/application.conf
```diff
@@ -138,3 +138,8 @@ urls {
   }
 }
 
+session-timeout{
+  signOutUrl = "https://www.gov.uk"
+  seconds = 1200 # 20 minutes
+  warning = 300  # 5 minutes (before timeout)
```
cmrhmrc, line 144: Might be useful to bring this timeout and the mongo session timeout together in a single config setting which could be overridden e.g. seconds = ${session.timeoutseconds}, with the mongodb setting similar. Units will need unifying i.e. both seconds or minutes

PR #72, opened 2020-04-30
EG-419,     Unify UI session timeout and session repository TTL

cmrhmrc:


test/mocks/MockAppConfig.scala
```diff
@@ -30,7 +30,7 @@ object MockAppConfig extends AppConfig {
   override val reportAProblemPartialUrl: String = "someUrl"
   override val reportAProblemNonJSUrl: String = "someJsUrl"
   override val externalGuidanceBaseUrl: String = "http://external-guidance-base-url"
-  override val sessionProcessTTLMinutes: Int = 20
+  override val sessionProcessTTLSeconds: Int = 1200
```
woollett-r, line 33: Should we use 900 here to make Mock configuration consistent with actual configuration?

macgrewal: @woollett-r as this used in unit tests it doesn't matter. The values are used to prove the code does what it should based on inputs, not that the inputs imitate real-life values. We have other tests for that.

Having said ☝️ we could just do away with the config value and `AppConfig.sessionProcessTTLSeconds` just read the `session-timeout.seconds` value thus removing the need to maintain 2 values in config.

Adding a comment in config that describes the dual usage of value should make it supportable also.

CC: @cmrhmrc

cmrhmrc: Updated as discussed in standup, ie removed sessionProcessTTLSeconds config an reworked Session repo to use session-timout.seconds config value

PR #73, opened 2020-05-04
EG-411 added feedback service identifier

RoheelArshad:


app/config/AppConfig.scala
```diff
@@ -46,7 +47,9 @@ class AppConfigImpl @Inject() (val config: Configuration, servicesConfig: Servic
   private val contactBaseUrl = servicesConfig.baseUrl("contact-frontend")
 
   private val assetsUrl = config.get[String]("assets.url")
-  private val serviceIdentifier = "MyService"
+  private val serviceIdentifier = "vegf"
```
cmrhmrc, line 50: Would remove private here as it may need to be referred to outside of the Config class

app/config/AppConfig.scala
```diff
@@ -46,7 +47,9 @@ class AppConfigImpl @Inject() (val config: Configuration, servicesConfig: Servic
...
+  lazy val host: String = servicesConfig.getString("host")
```
cmrhmrc, line 51: think this is part of EG-413 ??

conf/application.conf
```diff
@@ -143,3 +143,5 @@ session-timeout{
   seconds = 1200 # 20 minutes
   warning = 300  # 5 minutes (before timeout)
 }
+
+host = "http://localhost:9741"
```
cmrhmrc, line 147: part of EG-413 ??

PR #74, opened 2020-05-06
EG-412

RoheelArshad:


app/views/main_layout.scala.html
```diff
@@ -36,7 +36,7 @@
         tag = Some(Tag(
             content = Text("alpha")
         )),
-        content = HtmlContent("""This is a new service – your <a class="govuk-link" href="#">feedback</a> will help us to improve it.""")
+content = HtmlContent("This is a new service – your <a class='govuk-link' href=" + appConfig.feedbackUrl + ">feedback</a> will help us to improve it.")
```
cmrhmrc, line 39: pull display text from messages()

app/config/AppConfig.scala
```diff
@@ -56,6 +62,10 @@ class AppConfigImpl @Inject() (val config: Configuration, servicesConfig: Servic
   val languageMap: Map[String, Lang] = ListMap("english" -> Lang("en"), "cymraeg" -> Lang("cy"))
   lazy val sessionProcessTTLMinutes: Int = servicesConfig.getInt("mongodb.sessionProcessTTLMinutes")
 
+  private def requestUri(implicit request: RequestHeader) = ContinueUrl(host + request.uri).encodedUrl
+  def feedbackUrl(implicit request: RequestHeader): String =
```
cmrhmrc, line 66: This relative url may change, should be in configuration

app/config/AppConfig.scala
```diff
@@ -57,6 +60,12 @@ class AppConfigImpl @Inject() (val config: Configuration, servicesConfig: Servic
   val reportAProblemNonJSUrl: String = s"$contactBaseUrl/contact/problem_reports_nonjs?service=$serviceIdentifier"
   val languageMap: Map[String, Lang] = ListMap("english" -> Lang("en"), "cymraeg" -> Lang("cy"))
 
+
+  private def requestUri(implicit request: RequestHeader) = ContinueUrl(host + request.uri).encodedUrl
```
cmrhmrc, line 64: I would keep only the data from configuration here and leave the final construction of the url to its use in the main_layout where you have the request available , this applies to requestUri() and feedbackUrl()

PR #75, opened 2020-05-11
Eg 414

RoheelArshad:


app/views/main_layout.scala.html
```diff
@@ -58,6 +59,17 @@
     ))
 }
 
+@content = {
```
cmrhmrc, line 62: needs to be pulled from messages with welsh and english texts

app/views/main_layout.scala.html
```diff
@@ -58,6 +59,17 @@
...
+@contentBlock
```
cmrhmrc, line 63: useful to indent this to highlight how page made up of "contentBlock" parameter to layout, and the div you've added

app/views/main_layout.scala.html
```diff
@@ -58,6 +59,16 @@
     ))
 }
 
+@content = {
```
macgrewal, line 62: Use an implementation similar to https://github.com/hmrc/coronavirus-jrs-calculator-frontend/blob/master/app/views/templates/GovukLayoutWrapper.scala.html#L78-L86

PR #77, opened 2020-06-01
EG-529-1 added GTM scripts

RoheelArshad:


app/views/main_layout.scala.html
```diff
@@ -85,6 +85,11 @@
         message = Some(messages("timeout.dialog.message.text"))
         )
     )
+@appConfig.gtmContainer.map { containerId =>
```
LesWilson, line 88: minor, but can you indent new code to match code around it

RoheelArshad: Done

PR #78, opened 2020-06-01
EG-577, view replace ValueStanza with PageStanza

cmrhmrc:


woollett-r:
I've been through the pull request and it looks Ok to me.

PR #89, opened 2020-06-17
EG-621, ensure urls have leading /, blank and / only are invalid urls

cmrhmrc:


app/services/PageBuilder.scala
```diff
@@ -54,11 +54,11 @@ class PageBuilder extends ProcessPopulation {
     collectStanzas(key, Nil, Nil) match {
       case Right((ks, next, linked)) =>
         ks.head.stanza match {
-          case p: PageStanza if p.url.trim.isEmpty => Left(PageStanzaMissingOrUrlEmpty(ks.head.key))
+          case p: PageStanza if p.url.isEmpty || p.url.equals("/") => Left(PageUrlEmptyOrInvalid(ks.head.key))
```
woollett-r, line 57: You might want to use trim in the right hand test (p.url.trim.equals("/"). Actually you may need trim on the left too. From REPL "   ".isEmpty = false

cmrhmrc: Ah, I'd added trim within PageStanza when ensuring the leading /, i.e. the code which I removed. WIll add back

test/models/ocelot/ProcessJson.scala
```diff
@@ -39,9 +39,9 @@ trait ProcessJson {
       |  "contacts": [],
       |  "links": [],
       |  "flow": {
-      |    "1": {
+      |    "start": {
       |      "type": "PageStanza",
-      |      "url": "/feeling-bad",
+      |      "url": "feeling-bad",
```
woollett-r, line 44: The name "validOnePageJson" suggests this process should be valid, however, the url  "feeling-bad" is, I suspect, invalid. Is this correct?

test/services/UIBuilderSpec.scala
```diff
@@ -206,7 +206,7 @@ class UIBuilderSpec extends BaseSpec with ProcessJson {
     val questionWithAnswerHints: models.ocelot.stanzas.Question = Question(questionPhrase, answersWithHints, answerDestinations, false)
 
     val initialStanza = Seq(
-      ValueStanza(List(Value(Scalar, "PageUrl", "/")), Seq("1"), false),
+      ValueStanza(List(Value(Scalar, "PageUrl", "/blah")), Seq("1"), false),
```
woollett-r, line 209: You'll need to pull in the updates in master here.

cmrhmrc: will do

PR #97, opened 2020-06-22
EG-638, feedback links and footer translatability

cmrhmrc:


app/views/components/govukFooter.scala.html
```diff
@@ -0,0 +1,115 @@
...
+ *@
+
+@this()
+
+@(params: Footer = Footer())(implicit messages: Messages)
+
+@import uk.gov.hmrc.govukfrontend.views.Utils._
+@import play.twirl.api.HtmlFormat
+@import params._
+s
```
woollett-r, line 24: What is this for?

test/base/BaseSpec.scala
```diff
@@ -19,9 +19,16 @@ package base
...
 import org.scalatest.{MustMatchers, WordSpec}
 import play.api.libs.json._
+import play.twirl.api.Html
+import org.jsoup._
+import org.jsoup.nodes.{Document, Element}
+import scala.collection.JavaConverters._
 
 trait BaseSpec extends WordSpec with MustMatchers with ScalaFutures {
 
+  def asDocument(html: Html): Document = Jsoup.parse(html.toString)
```
woollett-r, line 29: I would have put these two methods in ViewSpec as their use is mainly confined to tests of views.

app/views/components/govukFooter.scala.html
cmrhmrc: :) its gone now

test/base/BaseSpec.scala
cmrhmrc: will move

conf/messages.cy
```diff
@@ -2,15 +2,39 @@ service.name = Welsh: External guidance
 hello.world = Helo o''r cais view-external-guide-frontend!
 hello.world.currentLanguage = Y gosodiad iaith cyfredol yw "{0}",
 
+service.phaseName = beta
```
woollett-r, line 5: Where possible I'd use the translations from the coronavirus project.

cmrhmrc: all in

PR #102, opened 2020-06-29
EG-650 Illegal character '{' in src attribute

woollett-r:


woollett-r:
Please check EG-649 as well as this seems to have been done.

PR #105, opened 2020-06-30
EG-654: Added support for SubSection Callout type and missing tests for Section

macgrewal:


macgrewal:
> Need to add
> case SubSection => Seq(H4(TextBuilder.fromPhrase(c.text))) to UiBuilder.fromCallout()
> case class H4(text: Text) extends UIComponent to Heading.scala
> case h: H4 => components.h4_heading(h) to render_components.scala.html
> and create an h4_heading.scala.html like h3_heading.scala.html

Yup, I have a separate task for that

PR #118, opened 2020-07-16
Lang fix

cmrhmrc:


app/controllers/SwitchLanguageController.scala
```diff
@@ -23,11 +23,27 @@ import config.AppConfig
 import uk.gov.hmrc.play.language.{LanguageController, LanguageUtils}
 
 @Singleton
-class SwitchLanguageController @Inject() (appConfig: AppConfig, languageUtils: LanguageUtils, cc: MessagesControllerComponents)
+class SwitchLanguageController @Inject(
+) (appConfig: AppConfig, languageUtils: LanguageUtils, cc: MessagesControllerComponents)
```
woollett-r, line 27: I'd move the closing parenthesis up to @Inject(

app/controllers/SwitchLanguageController.scala
```diff
@@ -23,11 +23,27 @@ import config.AppConfig
...
   override def languageMap: Map[String, Lang] = appConfig.languageMap
+  private val SwitchIndicatorKey = "switching-language"
+  private val FlashWithSwitchIndicator = Flash(Map(SwitchIndicatorKey -> "true"))
 
   // TODO requires suitable index like fallback URL
   override def fallbackURL: String = routes.AccessibilityStatementController.getPage().url
+
+  override def switchToLanguage(language: String): Action[AnyContent] = Action { implicit request =>
+    val enabled: Boolean = languageMap.get(language).exists(languageUtils.isLangAvailable)
+    val lang: Lang =
```
woollett-r, line 40: IntelliJ would like some braces in if statement

app/controllers/SwitchLanguageController.scala
cmrhmrc: :)

test/controllers/SwitchLanguageControllerSpec.scala
```diff
@@ -74,4 +75,18 @@ class SwitchLanguageControllerSpec extends WordSpec with Matchers with GuiceOneA
     }
   }
 
+  "Lang swtich" should {
+    "redirect to current page if referer is valid" in {
+      val result = controller.switchToLanguage("cy")(fakeRequest.withHeaders((HeaderNames.REFERER, s"${appConfig.baseHostUrl}/somepage")))
+      redirectLocation(result) shouldBe Some(s"${appConfig.baseHostUrl}/somepage")
+      confirmLangCookie(cookies(result), "en")
```
woollett-r, line 82: Are we moving to Welsh here? Should the cookie value be "cy"?

cmrhmrc: As its an override and likely to be fixed in the library, better to leave as much as found in order to see the changes. I think Intelij is wrong and the line itself would be better without the IF ;)

app/controllers/SwitchLanguageController.scala
cmrhmrc: As its an override and likely to be fixed in the library, better to leave as much as found in order to see the changes. I think Intelij is wrong and the line itself would be better without the IF ;)

app/config/AppConfig.scala
```diff
@@ -77,5 +78,6 @@ class AppConfigImpl @Inject() (val config: Configuration, servicesConfig: Servic
   lazy val timeoutInSeconds: Int = config.get[Int]("session-timeout.seconds")
   lazy val timeoutWarningInSeconds: Int = config.get[Int]("session-timeout.warning")
   lazy val baseUrl: String = config.get[String]("urls.baseUrl")
+  lazy val baseHostUrl: String = s"$host$baseUrl"
```
woollett-r, line 81: Update MockAppConfig for baseHostUrl

cmrhmrc: its there already??

app/controllers/SwitchLanguageController.scala
woollett-r: Ok

test/controllers/SwitchLanguageControllerSpec.scala
woollett-r: Lines removed

PR #119, opened 2020-07-18
EG-733, update aria-describedby

cmrhmrc:


LesWilson:
Compared local and staging using voice over and error was read out locally with the fix

PR #129, opened 2020-07-30
update config to make flash cookie secure

LesWilson:


LesWilson:
Have raised a PR in app-config-base

PR #141, opened 2020-08-13
EG-850 fix spacing issue when rendering an embedded link tag

LesWilson:


app/views/components/render_text.scala.html
```diff
@@ -18,16 +18,13 @@
 @import play.twirl.api.StringInterpolation
 
 @(txt: OcelotText)(implicit messages: Messages)
-
 @Html(
   txt.value(messages.lang).map{
-    case w: Words if w.bold => s"<strong>${w.toString}</strong>"
-    case w: Words => w.toString 
-    case l: Link => components.link(l).toString
+    case w: Words if w.bold => s"<strong>${w}</strong>"
```
woollett-r, line 23: This works because Word has a custom toString method that returns only the string component. If you are just using "w" you can remove the ${}.

PR #145, opened 2020-08-18
EG-874, added link-same and link-tab placeholder support

cmrhmrc:


test/services/BulletPointBuilderSpec.scala
```diff
@@ -44,63 +44,63 @@ class BulletPointBuilderSpec extends BaseSpec with ProcessJson {
 
       val text = ""
 
-      asString(BulletPointBuilder.fragmentsToDisplayAsSeq(text)) shouldBe text
+      asString(TextBuilder.flattenPlaceholders(text)) shouldBe text
```
woollett-r, line 47: It would be good to add some tests with link-same and link-tab to confirm parsing of new format links.

PR #148, opened 2020-08-24
EG-879, added InputStanza and InputType and tests

cmrhmrc:


test/models/ocelot/stanzas/InputStanzaSpec.scala
```diff
@@ -0,0 +1,119 @@
...
+    getStanzaJson("Number") -> expectedNumberStanza,
+    getStanzaJson("Text") -> expectedTextStanza,
+  )
+
+  jsonToStanzaMappings foreach { mapping =>
+    val (json, expectedStanza) = mapping
+    val jsonType = (json \ "ipt_type").as[String]
+
+    s"Reading valid JSON for a $jsonType Input" should {
+      s"create a ${expectedStanza.ipt_type} Callout Stanza" in {
```
woollett-r, line 57: Looks like a copy and paste error; Callout should be Input

test/models/ocelot/stanzas/InputStanzaSpec.scala
```diff
@@ -0,0 +1,119 @@
...
+  "Reading invalid JSON for a Input" should {
+    "generate a JsError" in {
+      getStanzaJson("invalid").validate[InputStanza] match {
+        case JsError(_) => succeed
+        case _ => fail("An instance of InputStanza should not be created when the note type is incorrect")
+      }
+    }
+  }
+
+  def toJsonFormat(typeName: String): String = s"""{"ipt_type":"$typeName","next":["1"],"name":0,"help":1,"label":"Price","placeholder":2,"stack":false}"""
```
woollett-r, line 73: Optional comment here. The method names could be more descriptive. It looks like the difference is between serialising an instance of InputStanza and its ancestor trait Stanza.

test/models/ocelot/stanzas/InputStanzaSpec.scala
cmrhmrc: Wiil chagne to Input

test/models/ocelot/stanzas/InputStanzaSpec.scala
cmrhmrc: done

PR #151, opened 2020-08-26
EG-880 Create Calculation stanza class and implement parsing

woollett-r:


app/models/ocelot/stanzas/CalcOperation.scala
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+package models.ocelot.stanzas
+
+import play.api.libs.functional.syntax._
+import play.api.libs.json._
+import play.api.libs.json.Reads._
+
+case class CalcOperation(left:String, calcOperationType: CalcOperationType, right: String, label: String)
```
cmrhmrc, line 23: you diverged (calcOperationType) from the naming inline with the json used on all other fields

PR #156, opened 2020-09-04
EG-909 Update viewer to use process code in process meta data in application urls

woollett-r:


app/models/ui/Page.scala
```diff
@@ -47,4 +47,4 @@ case class QuestionPage(val urlPath: String, question: Question) extends Page {
   val components: Seq[UIComponent] = Seq(question)
 }
 
-case class PageContext(page: Page, processStartUrl: Option[String], processTitle: Text, processId: String, answer: Option[String] = None)
+case class PageContext(page: Page, processStartUrl: Option[String], processTitle: Text, processCode: String, answer: Option[String] = None)
```
cmrhmrc, line 50: Might be aswell to leave processId in addition to processCode as we may need it at some point??

app/services/GuidanceService.scala
```diff
@@ -107,7 +109,7 @@ class GuidanceService @Inject() (
               .fold(err => {
                 logger.warn(s"Failed to parse process with error $err")
                 Left(InvalidProcessError)
-              }, pages => Right(pages.head.url))
+              }, pages => Right((pages.head.url, process.meta.processCode.getOrElse(process.meta.id))))
```
cmrhmrc, line 112: Cpuld add a method on Process to conditionally return the processCode or id depending on what is available (as titlePhrase), stops the spread of the uncertainty ;)

app/services/GuidanceService.scala
```diff
@@ -119,4 +121,15 @@ class GuidanceService @Inject() (
         Future.successful(Left(err))
     }
 
+  private def isCorrectProcess(process: Process, processCode: String ): Boolean = {
```
cmrhmrc, line 124: how about
process.meta.processCode.fold(process.meta.id == processCode)(pc => pc == processCode)

app/models/ui/Page.scala
woollett-r: Added processId back into PageContext.

app/services/GuidanceService.scala
woollett-r: Fixed

app/services/GuidanceService.scala
woollett-r: Fixed

app/models/ocelot/Process.scala
```diff
@@ -28,6 +28,8 @@ case class Process(meta: Meta, flow: Map[String, Stanza], phrases: Vector[Phrase
     titleIndex => phraseOption(titleIndex).getOrElse(Phrase(meta.title, meta.title))
   }
   lazy val startUrl: Option[String] = flow.get(Process.StartStanzaId).collect{case ps: PageStanza => ps.url}
+
```
cmrhmrc, line 31: lazy val as we want it to executed only once, probably only needed on Meta rather than both

app/models/ocelot/Meta.scala
```diff
@@ -19,7 +19,18 @@ package models.ocelot
...
+                title: String,
+                ocelot: Int,
+                lastAuthor: String,
+                lastUpdate: Long,
+                version: Int,
+                fileName: String,
+                titlePhrase: Option[Int] = None,
+                processCode: Option[String] = None) {
+
+  def getProcessIdentifier(): String = processCode.getOrElse(id)
```
cmrhmrc, line 32: lazy val as we only need it the check executed once

app/models/ocelot/Process.scala
woollett-r: I've removed this method from Process.

app/models/ocelot/Meta.scala
woollett-r: Added update

app/models/ocelot/Meta.scala
```diff
@@ -19,7 +19,18 @@ package models.ocelot
...
+                title: String,
+                ocelot: Int,
+                lastAuthor: String,
+                lastUpdate: Long,
+                version: Int,
+                fileName: String,
+                titlePhrase: Option[Int] = None,
+                processCode: Option[String] = None) {
+
+  lazy val getProcessIdentifier: String = processCode.getOrElse(id)
```
cmrhmrc, line 32: get implies a function rather an value

PR #169, opened 2020-09-23
EG-956, page builder supports Calculation and Choice stanzas, pages now graphs rather than sequences

cmrhmrc:


app/models/ocelot/stanzas/ValueStanza.scala
```diff
@@ -40,6 +40,7 @@ object Value {
 }
 
 case class ValueStanza(values: List[Value], override val next: Seq[String], stack: Boolean) extends Stanza {
+  override val visual: Boolean = false
```
LesWilson, line 43: As discussed, can we define a NonVisualStanza extending Stanza and setting visual = false.  Then any non-visual stanza extends this new stanza so they don't have to override visual attribute

app/models/ocelot/errors/GuidanceErrors.scala
```diff
@@ -37,6 +37,7 @@ case class PhraseNotFound(id: String, index: Int) extends FlowError
 case class LinkNotFound(id: String, index: Int) extends FlowError
 case class DuplicatePageUrl(id: String, url: String) extends FlowError
 case class MissingWelshText(id: String, index: String, english: String) extends FlowError
+case class VisualStanzasAfterQuestion(id: String) extends FlowError
```
LesWilson, line 40: Could we make this generic to cover all input types - so Question or new InputStanza being developed?  Guess depends if value in identifying it as a question or input stanza

cmrhmrc: You could include visual stanzas after input fields. Questions are a bit different as GDS makes the question text bethe H1 on the page and the visual stanzas before the question are wrapped in the question, so the question becomes the whole page and no visual stuff can follow.

test/services/PageBuilderSpec.scala
```diff
@@ -878,29 +847,33 @@ class PageBuilderSpec extends BaseSpec with ProcessJson with StanzaHelper {
...
 
-    expectedPageIds.zipWithIndex.foreach {
-      case (id, index) =>
-        indexedPages(index).id shouldBe id
-    }
+    pages.length shouldBe expectedPageIds.length
+
+    pages.forall(p => expectedPageIds.contains(p.id)) shouldBe true
+
+    // expectedPageIds.zipWithIndex.foreach {
```
LesWilson, line 856: can this be removed?

LesWilson (changes requested):
A few minor changes

app/models/ocelot/stanzas/ValueStanza.scala
cmrhmrc: will add

test/services/PageBuilderSpec.scala
cmrhmrc: yes will do

LesWilson (approved):

PR #170, opened 2020-09-23
EG-896 - updates to timeout handling

LesWilson:


conf/app.routes
```diff
@@ -5,6 +5,8 @@ GET   /language/:lang                    controllers.SwitchLanguageController.sw
 
 GET   /accessibility                     controllers.AccessibilityStatementController.getPage
 
+GET   /:processCode/session-timeout      controllers.SessionTimeoutPageController.getPage(processCode: String)
```
cmrhmrc, line 8: changed mind on this, /session-timeout/:processCode follows the accessibility out-of-guidance url pattern. Using /:processCode/session-timeout limits the possible urls within a piece of guidance unecessarilly.

app/controllers/SessionTimeoutPageController.scala
```diff
@@ -0,0 +1,141 @@
...
+              InternalServerError(errorHandler.internalServerErrorTemplateWithProcessCode(Some(processCode)))
+            }
+            case Right(_) => {
+              val title = Text(processContext.process.title.langs)
+              createDeleteYourAnswersResponse(title.asString(messages.lang), processCode)
+            }
+        }
+      }
+      case Left(NotFoundError) => {
+        // For some reason the timeout mechanism invokes the "getPage" method twice. In this case
```
cmrhmrc, line 90: I think with the conditional inclusion of the timeout dialog, this code is no longer required.

app/controllers/SessionTimeoutPageController.scala
```diff
@@ -0,0 +1,141 @@
...
+            Future.successful(createYourSessionHasExpiredResponse(messages("session.timeout.header.title"), processCode))
+          } else {
+            manageDeleteYourAnswersRequest(id, processCode)
+          }
+        }
+        case None => Future.successful(createYourSessionHasExpiredResponse(messages("session.timeout.header.title"), processCode))
+      }
+    }
+
+  def manageDeleteYourAnswersRequest(sessionId: SessionId, processCode: String)(implicit request: Request[_]): Future[Result] = {
```
cmrhmrc, line 67: would prefer to see this all this in the controller endpoint as it is the controllers main activity

app/controllers/SessionTimeoutPageController.scala
```diff
@@ -0,0 +1,141 @@
...
+
+    val logger = Logger(getClass)
+
+    def getPage(processCode: String): Action[AnyContent] = Action.async { implicit request =>
+
+      implicit val messages: Messages = mcc.messagesApi.preferred(request)
+
+      val sessionId: Option[SessionId] = hc.sessionId
+
+      sessionId match {
```
cmrhmrc, line 55: would have been useful to use the SessionFrontendController withExistingSession() here to get standard logging and use of common access to sessionId, however appreciate it would a bit of a re-write

app/controllers/SessionTimeoutPageController.scala
```diff
@@ -0,0 +1,141 @@
...
+    // Refresh cache to remove answers
+    implicit val messages: Messages = mcc.messagesApi.preferred(request)
+
+    service.getProcessContext(sessionId.value).flatMap {
+      case Right(processContext) if (processCode != processContext.process.meta.processCode) => {
+        logger.error(s"Unexpected process code encountered when removing session after timeout warning. Expected code $processCode; actual code ${processContext.process.meta.processCode}")
+        Future.successful(InternalServerError(errorHandler.internalServerErrorTemplate))
+      }
+      case Right(processContext) => {
+        service.removeSession(sessionId.value).map {
```
cmrhmrc, line 78: This isnt necessary as mongo will delete the session repo document automatically, what is important is to remove the current sessionId from the session

app/repositories/SessionRepository.scala
```diff
@@ -180,6 +184,32 @@ class DefaultSessionRepository @Inject() (config: AppConfig, component: Reactive
...
+      }
+    }.recover {
+      case lastError =>
+      logger.error(s"Error $lastError occurred in method get(key: String) attempting to retrieve session $key")
+      Left(DatabaseError)
+    }
+
+  }
+
+  def removeSession(key: String): Future[RequestOutcome[Unit]] = {
```
cmrhmrc, line 202: not needed as session repo documents are automatically deleted by mongo after session-timeout seconds

cmrhmrc, line 202: not needed as session repo documents are automatically deleted by mongo after session-timeout seconds

app/services/GuidanceService.scala
```diff
@@ -99,6 +101,8 @@ class GuidanceService @Inject() (
   Future[RequestOutcome[(String,String)]] =
     retrieveAndCache(processId, docId, connector.approvalProcess)
 
+  def removeSession(id: String): Future[RequestOutcome[Unit]] = sessionRepository.removeSession(id)
```
cmrhmrc, line 104: see above

conf/app.routes
LesWilson: Have posted on design channel to see which they prefer

app/controllers/SessionTimeoutPageController.scala
LesWilson: Think it is done this way because withExistingSession returns BadrequestError if no session and the code here is trying to distinguish between no session due to timeout or user hitting delete your answers as actions are different in each case

app/controllers/SessionTimeoutPageController.scala
LesWilson: moved in to main getPage

app/controllers/SessionTimeoutPageController.scala
LesWilson: removed

app/controllers/SessionTimeoutPageController.scala
LesWilson: removed

app/repositories/SessionRepository.scala
LesWilson: removed

app/services/GuidanceService.scala
LesWilson: removed

PR #173, opened 2020-09-30
EG-947, ChoiceStanza, Choice, ChoiceTests and unit tests

cmrhmrc:


app/models/ocelot/package.scala
```diff
@@ -21,9 +21,13 @@ import scala.util.matching.Regex
...
-  val labelRefRegex = s"\\[label:([0-9a-zA-Z\\s+]+)\\]".r
+  val labelRefRegex = s"\\[label:([0-9a-zA-Z\\s+_]+)\\]".r
+  val currencyRegex = "\\d+(\\.(\\d{2})?)?".r // From Ocelot
+
 
   def plSingleGroupCaptures(regex: Regex, str: String): List[String] = regex.findAllMatchIn(str).map(_.group(1)).toList
   def pageLinkIds(str: String): List[String] = plSingleGroupCaptures(pageLinkRegex, str)
   def labelReferences(str: String): List[String] = plSingleGroupCaptures(labelRefRegex, str)
+  def labelReference(str: String): Option[String] = plSingleGroupCaptures(labelRefRegex, str).headOption
+  def isCurrency(str: String): Boolean = currencyRegex.findFirstIn(str).fold(false)(_=> true)
```
woollett-r, line 32: This is an interesting case. Running the function in REPL the value 10.000000 is accepted as currency.

cmrhmrc: yes the Ocelot pattern allows many digits after the dp, could limit it, but kept as Ocelot

app/models/ocelot/stanzas/ChoiceStanza.scala
```diff
@@ -53,15 +49,65 @@ object ChoiceStanza {
...
     )(unlift(ChoiceStanza.unapply))
 
+}
+
+sealed trait ChoiceTest {
+  val left: String
+  val right: String
+  def eval(labels: Labels): Boolean
+  def value(arg: String, labels: Labels): String = labelReference(arg).fold(arg)(ref => labels.value(ref).getOrElse(""))
+  def op(f: (BigDecimal, BigDecimal) => Boolean, g: (String, String) => Boolean, labels: Labels): Boolean =
```
woollett-r, line 70: We may need a date function here eventually. Not certain about number; that may be handled by BigDecimal.

test/models/ocelot/LabelSpec.scala
```diff
@@ -45,4 +45,38 @@ class LabelSpec extends BaseSpec with ProcessJson {
 
   }
 
+  "LabelCache" must {
```
woollett-r, line 48: You could add a test for the interesting case of a label with an undefined value. If I'm right value will return Some().

cmrhmrc: Some("") yes, and non-existent None, will add test

test/models/ocelot/stanzas/ChoiceStanzaSpec.scala
```diff
@@ -187,11 +187,189 @@ class ChoiceStanzaSpec extends BaseSpec {
...
+                                                        ChoiceStanzaTest("3", LessThanOrEquals, "4"),
+                                                        ChoiceStanzaTest("3", NotEquals, "4")), false)
+      val choice = Choice(stanza)
+      val labels = Map("X"->Label("X", Some("33.5")), "Y"->Label("Y", Some("44")))
+      val lc = LabelCache(labels)
+      val expectedResult = (Seq("40"), lc)
+      choice.eval(lc) shouldBe expectedResult
+    }
+
+    "Evaluate to correct result when no tests succeed referencing labels" in {
```
woollett-r, line 231: This produces an interesting result. With three tests I think next should have four entries. The last one being used if all tests fail. I'm not certain what happens if there are three tests and three destinations and all of the tests fail.

test/models/ocelot/stanzas/ChoiceStanzaSpec.scala
```diff
@@ -187,11 +187,189 @@ class ChoiceStanzaSpec extends BaseSpec {
...
+    }
+
+    "provide support to MoreThanTest" in {
+      MoreThanTest("5", "5").eval(LabelCache()) shouldBe false
+
+      MoreThanTest("4", "5").eval(LabelCache()) shouldBe false
+
+      MoreThanTest("4", "3").eval(LabelCache()) shouldBe true
+
+      MoreThanTest("hello", "hello").eval(LabelCache()) shouldBe false
```
woollett-r, line 273: Are tests such as MoreThan appropriate for strings such as "hello". Should we be raising an error? What does Ocelot do?

test/models/ocelot/stanzas/ChoiceStanzaSpec.scala
cmrhmrc: Ah yes , test succeeds because the third of the ChoiceStanzaTests succeeds, if it had failed, it would have blown up because there arent 4 values in the next, will fix

test/models/ocelot/stanzas/ChoiceStanzaSpec.scala
cmrhmrc: It does the char/string equivalent as here i.e. a < b. I havent seen Ocelot generate any runtime errors, it works with what its given or does nothing as in the case of invalid or no input. If the guidance allows input as valid, dont think these tests should generate any runtime errors

woollett-r: As we discussed Scala probably follows the same approach as Java and Javascript when applying relational operators to strings.

PR #182, opened 2020-10-12
EG-946, merged page rendering etc

cmrhmrc:


app/controllers/GuidanceController.scala
```diff
@@ -94,6 +103,7 @@ class GuidanceController @Inject() (
             val pageContext = service.getPageContext(evalContext, Some(formData))
             pageContext.page match {
               case page: QuestionPage => Future.successful(BadRequest(questionView(page, pageContext, questionName(path), formWithErrors)))
+              case page: InputPage => Future.successful(BadRequest(inputView(page, pageContext, questionName(path), formWithErrors)))
               case _ => Future.successful(BadRequest(errorHandler.badRequestTemplateWithProcessCode(Some(processCode))))
             }
           },
```
woollett-r, line 109: Do we need a case for input page around line 120 also? I believe the form error section will handle errors caught by the Play framework. However, there are also errors from failed validation defined in the process.

app/models/ui/Input.scala
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.ui
+
+case class Input(text: Text, hint: Option[Text], body: Seq[UIComponent], errorMsgs: Seq[ErrorMsg] = Nil)
```
woollett-r, line 19: Should hint be optional? Looking at the UIBuilder hint is derived from "help" on the InputStanza and "help" is not optional.

cmrhmrc (commented):
Need to create a CurrencyInput stanza to allow currency specific validation etc

app/views/input_page.scala.html
```diff
@@ -0,0 +1,50 @@
...
+
+@layout(pageTitle(page), ctx.processTitle.asString(messages.lang), Some(ctx.processCode), ctx.processStartUrl, ctx.backLink) {
+
+  @components.error_summary(page.input.text, page.input.errorMsgs)
+
+  @formWithCSRF(action = controllers.routes.GuidanceController.submitPage(ctx.processCode, page.relativePath)) {
+
+    @components.input(page.input, inputName, form)
+
+    <button class="govuk-button" data-module="govuk-button">
```
woollett-r, line 42: Should we be using GovukButton here?

app/views/components/input_field.scala.html
```diff
@@ -0,0 +1,38 @@
...
+
+@(input: models.ui.Input, name: String, form: Form[_])(implicit request: Request[_], messages: Messages, labels: Labels)
+
+@optionalAria() = @{
+  Seq(input.hint.map(_ => s"${name}-item-hint"), input.errorMsgs.headOption.map(err => s"${err.id}-error")).flatten match {
+    case Nil => ""
+    case l => ariaDescribedBy(l.mkString(" "))
+  }
+}
+<div class="govuk-form-group">
```
woollett-r, line 29: There is an HMRC input frontend currency input component we could use here

app/controllers/GuidanceController.scala
woollett-r: Fixed

app/views/components/input_field.scala.html
woollett-r: However, it requires some text input whilst render_text generates Html. I guess this is why the components are written as they are.

PR #184, opened 2020-10-14
EG-1000, add DataInput validInput() and added to Question and Currenc…

cmrhmrc:
…yInput + tests and LessThan Choice test

### DIFF
--- a/test/models/ocelot/stanzas/ChoiceStanzaSpec.scala
+++ b/test/models/ocelot/stanzas/ChoiceStanzaSpec.scala
@@ -292,6 +292,18 @@ class ChoiceStanzaSpec extends BaseSpec {
       MoreThanOrEqualsTest("4", "hello").eval(LabelCache()) shouldBe false
     }
 
+    "provide support to LessThanTest" in {
+      LessThanTest("5", "5").eval(LabelCache()) shouldBe false
+
+      LessThanTest("4", "5").eval(LabelCache()) shouldBe true
+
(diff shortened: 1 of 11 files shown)


test/models/ocelot/stanzas/ChoiceStanzaSpec.scala
```diff
@@ -292,6 +292,18 @@ class ChoiceStanzaSpec extends BaseSpec {
       MoreThanOrEqualsTest("4", "hello").eval(LabelCache()) shouldBe false
     }
 
+    "provide support to LessThanTest" in {
```
woollett-r, line 295: All of this family of tests compare integers with integers, strings with strings or integers with strings. Given that IHT works with currency it would be good to add some more realistic tests.

test/models/ocelot/stanzas/InputStanzaSpec.scala
```diff
@@ -61,6 +62,66 @@ class InputStanzaSpec extends BaseSpec {
...
+        case currencyInput: CurrencyInput =>
+
+          currencyInput.validInput("33") shouldBe Some("33")
+        case _ => fail
+      }
+
+    }
+
+  }
+
```
woollett-r, line 124: I would add a test for valid value with two decimal places (i.e. 1000.00) and invalid with three decimal places (i.e. 1000.000) for completeness (maybe one and zero decimal places too)

PR #185, opened 2020-10-15
EG-1001, Controller validation, comma separated currency, fixes and tests, built on EG-1000

cmrhmrc:


app/models/ocelot/stanzas/QuestionStanza.scala
```diff
@@ -56,14 +56,13 @@ case class Question(text: Phrase,
...
-                }
-             }
-    else (None, labels)
+    validInput(value).fold[(Option[String], Labels)]((None, labels)){idx => {
+        val answerText = hintRegex.split(answers(idx.toInt).langs(0)).head.trim
+        (Some(next(idx.toInt)), label.fold(labels)(labels.update(_, answerText)))
+      }
+    }
+  def validInput(value: String): Option[String] =
+    asInt(value).fold[Option[String]](None)(idx => if (Range(0,answers.length).contains(idx)) Some(idx.toString) else None)
```
woollett-r, line 65: Apparently you can use answers.indices here.

PR #187, opened 2020-10-20
EG-1003 Update PageBuilder to manage row stanzas

woollett-r:


app/models/ocelot/stanzas/RowStanza.scala
```diff
@@ -39,3 +40,19 @@ object RowStanza {
 
     )(unlift(RowStanza.unapply))
 }
+
+case class Row( cells: Seq[Phrase],
+                override val next: Seq[String],
+                stack: Boolean = false,
+                override val links: List[String] = Nil) extends VisualStanza with Populated {
+
+  override val labelRefs: List[String] = if(cells.nonEmpty) cells.flatMap(c => labelReferences(c.langs(0))).toList else Nil
```
cmrhmrc, line 49: test is done twice here, can be `cells.toList.flatMap(labelReferences(_.langs(0))`

woollett-r: I've updated the code. However, I had to back off a bit on the code reduction.

PR #191, opened 2020-10-26
EG-1021, formatted label value output, for currency and default text, tests

cmrhmrc:


app/models/ocelot/package.scala
```diff
@@ -31,6 +32,7 @@ package object ocelot {
   def labelReferences(str: String): List[String] = plSingleGroupCaptures(labelRefRegex, str)
   def labelReference(str: String): Option[String] = plSingleGroupCaptures(labelRefRegex, str).headOption
   def isCurrency(str: String): Boolean = inputCurrencyRegex.findFirstIn(str).fold(false)(_ => true)
```
LesWilson, line 34: No test coverage for the isCurrency method

test/views/components/RenderTextSpec.scala
```diff
@@ -36,7 +36,7 @@ class RenderTextSpec extends WordSpec with Matchers with GuiceOneAppPerSuite {
 
   trait Test {
     implicit val labels: Labels = LabelCache(Map("Blah" -> Label("Blah", Some("a value")),
-                                                                "A-Label" -> Label("A-Label", Some("33.99"))))
+                                                                "A-Label" -> Label("A-Label", Some("33.9"))))
```
LesWilson, line 39: Could you add a label that has a value >= 1000 to test commas and an exact number to test that .00 added for currency, then add a couple of tests

LesWilson (changes requested):
Minor test updates

app/models/ocelot/package.scala
cmrhmrc: isCurrency no longer used, removed

test/views/components/RenderTextSpec.scala
cmrhmrc: yes, will do

LesWilson (approved):

PR #193, opened 2020-11-02
EG-1046 input accessibility

LesWilson:
Fixes based on WAVE accessibility reports - nee dto agree sizing of additional label input prompt


app/views/components/h1_heading_with_label.scala.html
```diff
@@ -0,0 +1,29 @@
+@*
```
LesWilson, line 1: This view may not be needed now we have changed how label is rendered

LesWilson:
replaced with new PR

PR #194, opened 2020-11-03
EG-916 Implement calculation stanza functionality for currency

woollett-r:


app/models/ocelot/stanzas/CalculationStanza.scala
```diff
@@ -41,4 +42,91 @@ object CalculationStanza {
...
+  val right: String
+  val label: String
+
+  def eval(labels: Labels) : Labels
+
+  def value(arg: String, labels: Labels): String = labelReference(arg).fold(arg){ref => labels.value(ref).getOrElse("")}
+
+  def addString(s1: String, s2: String) : Option[String] = Option(s1 + s2)
+
+  def subtractString(s1: String, s2: String) : Option[String] = {
```
cmrhmrc, line 61: This could be unsupportedOperation() and re-used elsewhere as it is likely, that most, if not all of the remaining calculation ops will be unsupported on non numeric data

PR #199, opened 2020-11-10
EG-972 "Your Decision" Callout stanza type for Confirmation Pages

woollett-r:


app/models/ui/ConfirmationPanel.scala
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.ui
+
+case class ConfirmationPanel (text: Text, additionalText: Seq[Text]) extends UIComponent
```
cmrhmrc, line 19: this could be a single Seq[Text], to allow for one or more lines

app/services/UIBuilder.scala
```diff
@@ -138,4 +139,42 @@ class UIBuilder {
       case x :: xs => partitionComponents(xs, errors, x +: others)
     }
 
+  private def fromSequenceWithLeadingCallout(sg: StackedGroup, calloutType: CalloutType, formData: Option[FormData])
```
cmrhmrc, line 142: would be simpler to make this YourCall specific. Callout in fn name should be Callouts as we know we have at least stacked callouts

app/services/UIBuilder.scala
```diff
@@ -138,4 +139,42 @@ class UIBuilder {
...
+    callouts.size match {
+      case 1 =>  fromStanzas(callouts.head +: stackStanzas(visualStanzas, Nil), Nil, formData)
+      case `groupSize` =>  Seq(fromYourCallGroup(callouts))
+      case _ => fromStanzas(StackedGroup(callouts) +: stackStanzas(visualStanzas, Nil), Nil, formData)
+    }
+
+  }
+
+  @tailrec
+  private def extractLeadingCallouts(calloutType: CalloutType,
```
cmrhmrc, line 158: this is equiv to
    val callouts: Seq[Callout] = sg.group.collect{case c: Callout if c.noteType == YourCall => c}
    val visualStanzas = sg.group.drop(callouts.length)

app/services/UIBuilder.scala
```diff
@@ -138,4 +139,42 @@ class UIBuilder {
...
+  private def fromSequenceWithLeadingCallout(sg: StackedGroup, calloutType: CalloutType, formData: Option[FormData])
+                              (implicit stanzaIdToUrlMap: Map[String, String]): Seq[UIComponent] = {
+
+    val (callouts, visualStanzas) = extractLeadingCallouts(calloutType, sg.group, Nil)
+
+    val groupSize: Int = sg.group.size
+
+    callouts.size match {
+      case 1 =>  fromStanzas(callouts.head +: stackStanzas(visualStanzas, Nil), Nil, formData)
+      case `groupSize` =>  Seq(fromYourCallGroup(callouts))
```
cmrhmrc, line 151: these two lines can replaced by the single loading the fromStanzas() accumulator with the processed YourCall stanzas as you know you have >1 stacked callouts 
_ => fromStanzas(stackStanzas(visualStanzas, Nil), Seq(fromYourCallGroup(callouts)), formData)

app/models/ui/ConfirmationPanel.scala
woollett-r: As ConfirmationPanel is a subclass of UIComponent it inherits "text". As the panel has a main title followed by additional lines of text it seemed reasonable to implement the class with "text" and "additionalText".

app/services/UIBuilder.scala
woollett-r: Originally it was a specific function for YourCall callouts, but I think there will be a case for other callout types to be handled in the same manner. As such I think this approach will be better in the long run. However, you are right about the name.

app/models/ui/ConfirmationPanel.scala
cmrhmrc: ok

app/services/UIBuilder.scala
woollett-r: Following  the removal of "extractLeadingCallouts", see below, I think it would be better to make the function specific to call outs of type yourCall.

app/services/UIBuilder.scala
woollett-r: The call to "extractLeadingCallouts" has been replaced by two fairly simple lines. Also as the last two parts of the case statement can be replaced by a single line and the first case is no longer possible The case statement has been removed also.

app/services/UIBuilder.scala
woollett-r: Applied in calling function.

PR #205, opened 2020-11-16
EG-1068, refactored Callout implementaion

cmrhmrc:


app/services/PageBuilder.scala
```diff
@@ -129,7 +129,7 @@ class PageBuilder extends ProcessPopulation {
   def fromPageDetails[A](pages: Seq[Page])(f: (String, String, String) => A): List[A] =
```
woollett-r, line 129: If this method is only used by PageBuilderSpec I would move it into the test class.

cmrhmrc: This is used in the backend (PageBuilder shared between viewer and backend)

test/models/ui/StackingSpec.scala
```diff
@@ -23,62 +23,63 @@ import models.ocelot.stanzas._
 class StackingSpec extends BaseSpec {
 
   trait Test {
-    def stackedCo(typ: CalloutType, stack: Boolean) = Callout(typ, Phrase(), Seq.empty, stack)
+    def stackedTitle(stack: Boolean) = TitleCallout(Phrase(), Seq.empty, stack)
```
woollett-r, line 26: Can you add the type on these two methods to shut up Intellij please?

cmrhmrc: done

PR #213, opened 2020-11-27
EG-1111, forced all table RowStanza groups to have a caption

cmrhmrc:


LesWilson (approved):
"tables" now render as per Adam's update

PR #216, opened 2020-12-01
EG-1033-Exp Fix for back link on viewer

woollett-r:


woollett-r:
The journey tests will need to be updated before this PR can be merged.

PR #222, opened 2020-12-07
EG-1134 date input

LesWilson:
Merged in current master changes

### DIFF
--- a/app/controllers/GuidanceController.scala
+++ b/app/controllers/GuidanceController.scala
@@ -20,20 +20,21 @@ import config.{AppConfig, ErrorHandler}
 import javax.inject.{Inject, Singleton}
 import play.api.i18n.Messages
 import play.api.mvc._
-import play.api.data.{FormError, Form}
+import play.api.data.{Form, FormError}
 import services.GuidanceService
 import uk.gov.hmrc.play.bootstrap.controller.FrontendController
 import models.errors._
 import models.{PageContext, PageEvaluationContext}
-import models.ui.{StandardPage, InputPage, QuestionPage, FormData}
+import models.ui.{DateInputPage, FormData, InputPage, QuestionPage, StandardPage}
 import models.ocelot.stanzas.DataInput
 import forms.SubmittedAnswerFormProvider
-import views.html.{input_page, standard_page, question_page}
+import views.html.{input_date_page, input_page, question_page, standard_page}
 import play.api.Logger
-import scala.concurrent.Future
+
 import scala.concurrent.ExecutionContext.Implicits.global
 import controllers.actions.SessionIdAction
 import play.twirl.api.Html
+
 import scala.concurrent.Future
 import models.ocelot.KeyedStanza
 
@@ -45,13 +46,14 @@ class GuidanceController @Inject() (
     standardView: standard_page,
     questionView: question_page,
     inputView: input_page,
+    dateInputView: input_date_page,
     formProvider: SubmittedAnswerFormProvider,
     service: GuidanceService,
     mcc: MessagesControllerComponents
 ) extends FrontendController(mcc)
   with SessionFrontendController {
 
-  val logger = Logger(getClass)
+  val logger: Logger = Logger(getClass)
 
   def getPage(processCode: String, path: String, p: Option[String]): Action[AnyContent] = sessionIdAction.async { implicit request =>
     implicit val messages: Messages = mcc.messagesApi.preferred(request)
@@ -69,6 +71,8 @@ class GuidanceController @Inject() (
             Future.successful(Ok(questionView(page, pageContext, questionName(path), populatedForm(pageContext, path))))
           case page: InputPage =>
             Future.successful(Ok(inputView(page, pageContext, questionName(path), populatedForm(pageContext, path))))
+          case page: DateInputPage =>
+            Future.successful(Ok(dateInputView(page, pageContext, questionName(path), populatedForm(pageContext, path))))
         }
       case Left(NotFoundError) =>
         logger.warn(s"Request for PageContext at /$path returned NotFound, returning NotFound")
--- a/app/controllers/GuidanceController.scala
+++ b/app/controllers/GuidanceController.scala
@@ -56,24 +56,9 @@ class GuidanceController @Inject() (
   val logger: Logger = Logger(getClass)
 
   def getPage(processCode: String, path: String, p: Option[String]): Action[AnyContent] = sessionIdAction.async { implicit request =>
-    implicit val messages: Messages = mcc.messagesApi.preferred(request)
     withExistingSession[PageContext](service.getPageContext(processCode, s"/$path", p.isDefined, _)).flatMap {
       case Right(pageContext) =>
-        logger.info(s"Retrieved page at ${pageContext.page.urlPath}, start at ${pageContext.processStartUrl}," +
-                    s" answer = ${pageContext.answer}, backLink = ${pageContext.backLink}")
-        pageContext.page match {
-          case page: StandardPage =>
-            service.saveLabels(pageContext.sessionId, pageContext.labels).map{
-              case Right(_) => Ok(standardView(page, pageContext))
-              case Left(err) => InternalServerError(errorHandler.internalServerErrorTemplate)
-            }
-          case page: QuestionPage =>
-            Future.successful(Ok(questionView(page, pageContext, questionName(path), populatedForm(pageContext, path))))
-          case page: InputPage =>
-            Future.successful(Ok(inputView(page, pageContext, questionName(path), populatedForm(pageContext, path))))
-          case page: DateInputPage =>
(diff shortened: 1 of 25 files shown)


app/services/PageBuilder.scala
```diff
@@ -133,6 +133,8 @@ class PageBuilder extends ProcessPopulation {
           f(page.id, page.url, text.langs(0))
```
woollett-r, line 133: The methods fromPageDetails and pagesWithValidation are used only in the test classes which makes me wonder if they should be relocated into the test classes.

cmrhmrc: Richard, fromPageDetails and pagesWithValidation are both used in EG and this code is shared between both the viewer and EG

app/controllers/GuidanceController.scala
```diff
@@ -45,13 +46,14 @@ class GuidanceController @Inject() (
...
     formProvider: SubmittedAnswerFormProvider,
     service: GuidanceService,
     mcc: MessagesControllerComponents
 ) extends FrontendController(mcc)
   with SessionFrontendController {
 
-  val logger = Logger(getClass)
+  val logger: Logger = Logger(getClass)
 
   def getPage(processCode: String, path: String, p: Option[String]): Action[AnyContent] = sessionIdAction.async { implicit request =>
```
woollett-r, line 58: This method is exceeding the cyclomatic complexity threshold. I'd be tempted to move the code handling a successful data retrieval (lines 62 to 76 to a private method). I have done something similar for the submission code in EG-1137-Exp.

app/views/components/input_date.scala.html
```diff
@@ -0,0 +1,55 @@
...
+      @components.h1_heading_with_label(H1(input.text), name + "-0")
+      @{input.hint.map(hintText => <span id="input-hint" class="govuk-hint">{components.render_text(hintText)}</span>)}
+      @components.error_message(input.errorMsgs)
+      @components.input_date_field(input, name, form)
+
+  } else {
+    @components.h1_heading(H1(input.text))
+    @components.render_components(input.body)
+
+    <fieldset class="govuk-fieldset" @optionalAria()>
```
woollett-r, line 42: Why is the fieldset element only used when there are additional components on the page?

LesWilson: That was what was thrown up in the accessibility review to remove it from single input - but matches the GDS style guide for date input - a fieldset groups fields which are related

app/views/components/input_date_field.scala.html
```diff
@@ -0,0 +1,57 @@
...
+
+@(input: models.ui.DateInput, name: String, form: Form[_])(implicit request: Request[_], messages: Messages, ctx: PageContext)
+
+@optionalAria() = @{
+  Seq(input.hint.map(_ => s"$name-item-hint"), input.errorMsgs.headOption.map(err => s"${err.id}-error")).flatten match {
+    case Nil => ""
+    case l => ariaDescribedBy("input-hint")
+  }
+}
+<div class="govuk-date-input" id="date-input">
```
woollett-r, line 25: The id here is not going to be unique if there are multiple dates on a form. Further, if there are multiple components on a page a label is created for the id "name-0", should this be the id?

app/views/components/input_date_field.scala.html
```diff
@@ -0,0 +1,57 @@
...
+    <div class="govuk-date-input__item">
+        <div class="govuk-form-group">
+            <label class="govuk-label govuk-date-input__label" for="day">
+              @messages("label.day")
+            </label>
+            <input class="govuk-input govuk-date-input__input govuk-input--width-2" id="day" name="day" value="@form("day").value" type="text" pattern="[0-9]*" inputmode="numeric">
```
woollett-r, line 31: Would it be better to use @name-day here?

app/views/components/input_date_field.scala.html
```diff
@@ -0,0 +1,57 @@
...
+    </div>
+    <div class="govuk-date-input__item">
+        <div class="govuk-form-group">
+            <label class="govuk-label govuk-date-input__label" for="year">
+               @messages("label.year")
+            </label>
+            <input class="govuk-input govuk-date-input__input govuk-input--width-4" id="year" name="year" value="@form("year").value" type="text" pattern="[0-9]*" inputmode="numeric"></div>
+    </div>
+</div>
+@*<div class="govuk-form-group">*@
```
woollett-r, line 49: If these lines are commented out they should be removed.

app/views/components/input_date.scala.html
woollett-r: Ok

app/services/PageBuilder.scala
woollett-r: Ok, I can one usage in external-guidance.

app/controllers/GuidanceController.scala
LesWilson: will take a look

app/views/components/input_date_field.scala.html
LesWilson: As this assumes one input per page, adding -0 at end of id doesn't really make much sense.  The date fields are called day, month and year, so would have to change if we have multiple dates on a page - but this woud be the least of the changes compared to controller side of things.

app/views/components/input_date_field.scala.html
LesWilson: see above, not sure why we need the @name as we will have a specific date handler in back end

app/views/components/input_date_field.scala.html
LesWilson: yes, will sort out

app/views/components/input_date_field.scala.html
woollett-r: Ok

app/views/components/input_date_field.scala.html
woollett-r: Yes, it all depends on how many input components we have on a page. If it is only one this is ok.

test/controllers/GuidanceControllerSpec.scala
```diff
@@ -27,15 +28,16 @@ import play.api.test.Helpers._
 import uk.gov.hmrc.play.bootstrap.tools.Stubs.stubMessagesControllerComponents
 import uk.gov.hmrc.http.SessionKeys
 import forms.SubmittedAnswerFormProvider
-import models.{PageEvaluationContext, PageContext}
-import models.ocelot.{Process, ProcessJson, Phrase,KeyedStanza, Page => OcelotPage}
-import models.ocelot.stanzas.{Question => OcelotQuestion,_}
-import models.ui.{CurrencyInput => UiCurrencyInput,_}
+import models.{PageContext, PageEvaluationContext}
+import models.ocelot.{KeyedStanza, Phrase, Process, ProcessJson, Page => OcelotPage}
```
woollett-r, line 32: Rearranging the order to alphabetic order; that's the kind of OCD update I do!

LesWilson: Thats the sort of OCD update Intellij does :)

test/views/components/InputSpec.scala
```diff
@@ -153,4 +153,95 @@ class InputSpec extends WordSpec with Matchers with GuiceOneAppPerSuite {
       }
    }
   }
+
+  "English Date Input component" must {
```
woollett-r, line 157: For completeness I'd add a test for the existence of a fieldset when there are additional components on the page.

app/views/components/h1_heading_with_fieldset.scala.html
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import ui.H1
+
+@( h1: H1 )( implicit messages: Messages, ctx: PageContext)
+<legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
```
woollett-r, line 20: xl seems on the large size. What is your reasoning?

woollett-r: I see question uses xl so I guess it is consistent.

LesWilson: It matched the input with label component

cmrhmrc: have a look at h1_heading.scala.html. There is a setting on the page which determines if the reduced heading sizes should be used. something like class="@if(ctx.page.reduceHeadings){govuk-heading-l} else {govuk-heading-xl}"

cmrhmrc: also applies to h1_heading_with_label.scala.html

PR #223, opened 2020-12-10
EG-1147, refactor of table and summary list support in line with accessibilty report

cmrhmrc:
Needs journey tests EG-1147 to be merged first.


test/services/UIBuilderSpec.scala
```diff
@@ -478,76 +480,57 @@ class UIBuilderSpec extends BaseSpec with ProcessJson {
...
-        case Seq(_: H1, Table(_, None, Seq())) => succeed
+        case Seq(_: H1, _: NameValueSummaryList) => succeed
         case x => fail(s"Found $x")
       }
     }
 
     "Convert a non-summarylist RowGroup into a table with a heading line" in new TableTest {
-      val p = uiBuilder.buildPage("/start", Seq(TitleCallout(headingPhrase, Seq.empty, false),
-                                                captionRowGroup))
+      println(s"IsCandidate ${tableRowGroup.isTableCandidate}")
```
woollett-r, line 502: Remove println

test/services/UIBuilderSpec.scala
```diff
@@ -478,76 +480,57 @@ class UIBuilderSpec extends BaseSpec with ProcessJson {
...
+      val p = uiBuilder.buildPage("/start", Seq(SubSectionCallout(headingPhrase, Seq.empty, false),
+                                                tableRowGroup))
       p.components match {
-        case Seq(_: H1, Table(Text(Seq(Words("HELLO", true)), Seq(Words("HELLO", true))), None, _)) => succeed
+        case Seq(Table(Text(headingPhrase,_), tableHeading, _)) => succeed
         case x => fail(s"Found $x")
       }
     }
 
     "convert a non-summarylist RowGroup stacked to a SubSection into a table with caption and a heading" in new TableTest {
```
woollett-r, line 511: I think stack is already true. The RowGroup stack value is set to the value of stack for the first row. If you make the RowGroup stack false the previous test will fail.

PR #224, opened 2020-12-10
EG-1152 Apply fix for back link after submission redirect to previous page

woollett-r:


app/controllers/GuidanceController.scala
```diff
@@ -148,4 +151,12 @@ class GuidanceController @Inject() (
   }
 
   private def questionName(path: String): String = path.reverse.takeWhile(_ != '/').reverse
+
+  private def previousPageQueryString(url: String, backLink: Option[String]): Option[String] =
```
cmrhmrc, line 155: good example for collect()

backLink.collect{case bl if bl == url => 1}

PR #227, opened 2020-12-16
EG-1151, Outlaw page redirects except after input, enable return to page id on input error

cmrhmrc:


test/services/PageBuilderSpec.scala
```diff
@@ -122,6 +122,34 @@ class PageBuilderSpec extends BaseSpec with ProcessJson with StanzaHelper {
...
 
+    "detect SharedDataInputStanza error" in {
+      val flow = Map(
+        Process.StartStanzaId -> PageStanza("/url", Seq("1"), true),
+        "1" -> InstructionStanza(0, Seq("2"), None, false),
+        "2" -> InputStanza(Currency, Seq("4"), 0, Some(0), "Label", None, false),
+        "4" -> Choice(ChoiceStanza(Seq("5","end"), Seq(ChoiceStanzaTest("[label:label]", LessThanOrEquals, "8")), false)),
+        "5" -> PageStanza("/url2", Seq("1"), true),
+        "6" -> InstructionStanza(0, Seq("2"), None, false),
+        "2" -> InputStanza(Currency, Seq("4"), 0, Some(0), "Label", None, false),
```
LesWilson, line 133: Can you have the same id ("2") repeated from Ocelot?

cmrhmrc: This is what happens when 2 pages link to the same shared stanzas. Each built page gets a copy of the shared stanzas in their list of stanzas. In Ocelot there is no copy, but their are one or more stanzas which are linked to 2 or more separate pages.

LesWilson (approved):

PR #228, opened 2020-12-21
EG-1169, Added FormPage, Split viewer errors into Required and Type

cmrhmrc:


app/controllers/GuidanceController.scala
```diff
@@ -54,9 +52,19 @@ class GuidanceController @Inject() (
...
     withExistingSession[PageContext](service.getPageContext(processCode, s"/$path", p.isDefined, _)).flatMap {
-      case Right(pageContext) =>
-        showPage(path, pageContext)
+      case Right(pageCtx) =>
+        logger.info(s"Retrieved page at ${pageCtx.page.urlPath}, start at ${pageCtx.processStartUrl}," +
+                    s" answer = ${pageCtx.answer}, backLink = ${pageCtx.backLink}")
+        pageCtx.page match {
+          case page: StandardPage => service.saveLabels(pageCtx.sessionId, pageCtx.labels).map {
+              case Right(_) => Ok(standardView(page, pageCtx))
+              case Left(err) => InternalServerError(errorHandler.internalServerErrorTemplate)
```
LesWilson, line 63: err not used

app/services/GuidanceService.scala
```diff
@@ -59,21 +58,27 @@ class GuidanceService @Inject() (
               logger.error(s"PageBuilder error $err on process ${process.meta.id} with sessionId $sessionId")
               Left(InvalidProcessError)
             },
-            page =>
+            page => {
+              val stanzaIdToUrlMap = urlToPageId.map{case (k, v) => (v, s"${appConfig.baseUrl}/${processCode}${k}")}.toMap
```
LesWilson, line 62: don't need toMap

app/controllers/GuidanceController.scala
cmrhmrc: replace with _

app/services/GuidanceService.scala
cmrhmrc: removed

LesWilson (approved):
Looks good :)

PR #231, opened 2020-12-23
EG-1160, date input validation with converstion to LocalDate and back to standard string format

cmrhmrc:
Also add optional DataInput component to PageContext. Present on form page and None on standard pages


cmrhmrc:
Note code branch name is EG-1160-fixedformat not EG-1160

LesWilson (approved):

PR #236, opened 2021-01-05
EG-1179, Text input

cmrhmrc:
Sorry many files changed. Many changes are as a result of adding the visual stanzas to the PageContext which results in changes to the tests

### DIFF
--- a/test/models/ocelot/OcelotPackageSpec.scala
+++ b/test/models/ocelot/OcelotPackageSpec.scala
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2020 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.ocelot
+
+import base.BaseSpec
+import org.scalatest.Inspectors.forAll
+
+class OcelotPackageSpec extends BaseSpec {
+
+  "Date conversion" must {
+    "recognise a valid date" in {
+      val validDates: List[String] = List("30/04/2000", "3/12/2000", "13/4/2000", "31/3/2130")
+      forAll(validDates) { validDate =>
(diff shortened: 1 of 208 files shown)


test/models/ocelot/OcelotPackageSpec.scala
```diff
@@ -89,6 +89,25 @@ class OcelotPackageSpec extends BaseSpec {
     }
   }
 
+  "Text conversion" must {
+    "recognise a valid text" in {
+      val validNumbers: List[String] = List("Hello", "World", "030")
```
LesWilson, line 94: can we change the name of this variable

test/views/components/InputSpec.scala
```diff
@@ -75,14 +75,99 @@ class InputSpec extends WordSpec with Matchers with GuiceOneAppPerSuite {
...
+  "English Text Input component" must {
+
+    trait TextTest extends Test {
+      override val input: TextInput = TextInput(Text(i1), Some(Text(i1Hint)), Seq(h2, h3, h4, para1))
+
+      override val inputWithoutBody: TextInput = TextInput(Text(i1), None, Seq.empty)
+      override val inputWithHintAndNoBody: TextInput = TextInput(Text(i1), Some(Text(i1Hint)), Seq.empty)
+      override val inputWithHintAndErrors: TextInput = TextInput(Text(i1), Some(Text(i1Hint)), Seq(para1), Seq(errorMsg))
+      val textInput: TextInput = models.ui.TextInput(Text(), None, Seq.empty)
+      val textInputPage: FormPage = models.ui.FormPage("/url", textInput)
```
LesWilson, line 168: Do we need this TextInput - can we just use the input defined above?
Can we remove the models.ui. from these definitions

LesWilson (commented):
a couple of minor changes.  
Other changes we talked about are more cosmetic and we can discuss more when we have time to tidy things up around PageContext creation to reduce impact of future signature changes

test/views/components/InputSpec.scala
cmrhmrc: They look similar, but the one above uses CurrencyInput as the Input implementation, so itsthe same tests but run through with the TextInput type

test/models/ocelot/OcelotPackageSpec.scala
cmrhmrc: will do

test/views/components/InputSpec.scala
LesWilson: Sorry, the picture above doesn't show the input I meant - there is a TextInput defined at the top of this section that is used in tests

cmrhmrc: The one above is unused, deleted now

LesWilson (approved):

PR #237, opened 2021-01-06
EG-1186 Synchronise date string to Java date conversion

woollett-r:


woollett-r:
Update already applied in another update.

PR #248, opened 2021-01-25
EG-1197 Date feature - support multiple tests

woollett-r:


app/core/models/ocelot/stanzas/ChoiceStanza.scala
```diff
@@ -67,46 +70,61 @@ sealed trait ChoiceTest {
   val right: String
   def eval(labels: Labels): Boolean
   def value(arg: String, labels: Labels): String = labelReference(arg).fold(arg)(ref => labels.value(ref).getOrElse(""))
-  def op(f: (BigDecimal, BigDecimal) => Boolean, g: (String, String) => Boolean, labels: Labels): Boolean = {
+  def moreThanOrEquals(date: LocalDate, other: LocalDate): Boolean = date.compareTo(other) >= 0
```
cmrhmrc, line 73: think these fns dont belong here, the ChoiceTest trait is common to all tests without any type or test specific fns, suggest using an anonymous fn as other tests i.e. _.compareTo(_) >= 0 and _.compareTo(_) <= 0

test/core/models/ocelot/stanzas/ChoiceStanzaSpec.scala
```diff
@@ -144,7 +144,7 @@ class ChoiceStanzaSpec extends BaseSpec {
 
   "ChoiceStanza" must {
 
-    "deserialise from json" in {
+    "deserialize from json" in {
```
cmrhmrc, line 147: American spelling ????

app/core/models/ocelot/stanzas/ChoiceStanza.scala
woollett-r: Thanks

test/core/models/ocelot/stanzas/ChoiceStanzaSpec.scala
woollett-r: Surprisingly, according to the Shorter Oxford Dictionary (1983) no. The correct spelling is serialize. I just changed the spelling because IntelliJ flagged up a mistake. Now that I am better informed I've updated the rest of the class.

It may be that the root is Greek not Latin like authorize.

PR #250, opened 2021-01-27
EG-1235, refactor passphrase handling, allow admin viewing of unprotected passphrase guidance pages

cmrhmrc:


app/services/GuidanceService.scala
```diff
@@ -175,5 +179,12 @@ class GuidanceService @Inject() (
         )
     }
 
-  private def isAuthenticationUrl(url: String): Boolean = url.drop(1).equals(core.models.ocelot.Process.SecuredProcessStartUrl)
+  private def isAuthenticationUrl(url: String): Boolean = url.drop(1).equals(SecuredProcess.SecuredProcessStartUrl)
+
+  private def map[A](f: Retrieve[A])(g: A => A)(implicit ec: ExecutionContext): Retrieve[A] = {
```
woollett-r, line 184: It may be appropriate, but map is a confusing name for this method.

cmrhmrc: its a standard implimentation of map(), as you'd find on List , Option etc, so intentionally called map() to indicate this

woollett-r: It is certainly an interesting function.

test/core/models/ocelot/MetaSpec.scala
```diff
@@ -79,7 +79,7 @@ class MetaSpec extends BaseSpec {
       result shouldBe validModelWithOptionalProperties
     }
 
-    missingJsObjectAttrTests[Meta](validJsonWithOptionalProperties, List("titlePhrase", "processCode", "version"))
+    missingJsObjectAttrTests[Meta](validJsonWithOptionalProperties, List("titlePhrase", "processCode", "version", "lastAuthor", "lastUpdate", "filename"))
```
woollett-r, line 82: As with the backend "processCode" is a mandatory attribute and as such should not be excluded from the property tests.

test/services/SecuredProcessBuilderSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+  def injector: Injector = app.injector
+  val messagesApi: MessagesApi = app.injector.instanceOf[MessagesApi]
+
+  val meta: Meta = Json.parse(prototypeMetaSection).as[Meta]
+  val flow: Map[String, Stanza] = Json.parse(prototypeFlowSection).as[Map[String, Stanza]]
+  val phrases: Vector[Phrase] = Json.parse(prototypePhrasesSection).as[Vector[Phrase]]
+  val links: Vector[Link] = Json.parse(prototypeLinksSection).as[Vector[Link]]
+
+  val process: Process = prototypeJson.as[Process]
+  val passphraseProcess: Process = validOnePageProcessWithPassPhrase.as[Process]
```
woollett-r, line 38: These two values look surprisingly similar.

cmrhmrc: removed

PR #256, opened 2021-02-09
EG-1269, timescale placeholder

cmrhmrc:


app/core/services/Placeholders.scala
```diff
@@ -0,0 +1,61 @@
...
+@Singleton
+class Placeholders {
+  val TaxYearStartDay: Int = 6
+  val TaxYearStartMonth: Int = 4
+  val timescalePattern = "\\[timescale:(today|CY([\\-+]\\d+)?)(:(long|short))?\\]"
+  val timescaleRegex: Regex = s"${timescalePattern}".r
+
+  def long(date: LocalDate): Int = date.getYear
+  def short(date: LocalDate): Int = long(date) % 100
+  def CY(offset: Int, when: LocalDate): LocalDate = {
```
woollett-r, line 38: I have my doubts about this method.

app/core/services/Placeholders.scala
```diff
@@ -0,0 +1,61 @@
...
+    val candidateStartDate = when.withMonth(TaxYearStartMonth).withDayOfMonth(TaxYearStartDay)
+    candidateStartDate.minusYears(-offset + (if (when.isAfter(candidateStartDate)) 0 else 1))
+  }
+
+  private val TodayOrCyGroup: Int = 1
+  private val CyOffsetGroup: Int = 2
```
woollett-r, line 44: What does group 3 match?

app/core/services/PageBuilder.scala
```diff
@@ -16,15 +16,15 @@
...
+import javax.inject.{Inject, Singleton}
 import core.models.ocelot._
 import core.models.ocelot.stanzas._
 import core.models.ocelot.errors._
 import play.api.Logger
 import scala.annotation.tailrec
 
 @Singleton
-class PageBuilder extends ProcessPopulation {
+class PageBuilder @Inject() (val placeholders: Placeholders) extends ProcessPopulation with PlaceholderProvider {
```
woollett-r, line 27: Is it necessary to inject an instance of PlaceHolders as it is not being used here.

app/core/services/Placeholders.scala
cmrhmrc: fixed bug when "when" first day of tax year

app/core/services/Placeholders.scala
cmrhmrc: Its a unused match used around the long/short modifier i.e. when using long it would be ":long"

app/core/services/PageBuilder.scala
cmrhmrc: yes, ProcessPopulation expects whatever it is mixed into to be a PlaceholderProvider

app/core/services/ProcessPopulation.scala
```diff
@@ -22,6 +22,7 @@ import core.models.ocelot.errors._
 import scala.annotation.tailrec
 
 trait ProcessPopulation {
+  this: PlaceholderProvider =>
```
woollett-r, line 25: What does this do?

cmrhmrc: means "this" has to be a PlaceholderProvider i.e. ProcessPopulation can only be mixed into a class which implements PlaceholderProvider

woollett-r: Thanks you. Also helps explain why PageBuilder needs an instance of PlaceHolders.

test/core/services/PlaceholdersSpec.scala
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+package core.services
+
+import base.BaseSpec
+import java.time.LocalDate
+
+class PlaceholdersSpec extends BaseSpec {
+  val pls: Placeholders = new Placeholders
+  val today: LocalDate = LocalDate.of(2020, 6, 24)
```
woollett-r, line 40: Today gives us a date in the main part of the tax year (i.e. April to December). For completeness it would be good to replicate the tests with a date between Jan and March inclusive.

cmrhmrc: will add

PR #259, opened 2021-02-17
EG-1290 Calculation stanza - Build list in the value stanza

woollett-r:


app/core/models/ocelot/stanzas/CalculationStanza.scala
```diff
@@ -63,21 +65,43 @@ sealed trait Operation {
     labels.value(ref).getOrElse("")
   }
 
-  def unsupportedOperation(operationName: String)(arg1: Any, arg2: Any): Option[String] = {
+  def valueAsOption(arg: String, labels: Labels): Option[String] = labelReference(arg).fold[Option[String]](Some(arg)){ref =>
```
cmrhmrc, line 68: think value and valueAsOption could be replaced by a single fn value returning an Option[String]

app/core/models/ocelot/stanzas/CalculationStanza.scala
```diff
@@ -63,21 +65,43 @@ sealed trait Operation {
...
+    labels.value(ref)
+  }
+
+  def valueAsListOption(arg: String, labels: Labels): Option[List[String]] = {
```
cmrhmrc, line 72: valueAsListOption accessing a list label and returning as a list, wouldn't listValue be clearer as valueAsList suggests this is a conversion fn. listValue would also match value (in comment above)

app/core/models/ocelot/stanzas/CalculationStanza.scala
woollett-r: Done

app/core/models/ocelot/stanzas/CalculationStanza.scala
```diff
@@ -125,16 +158,54 @@ sealed trait Operation {
 
 case class AddOperation(left: String, right: String, label: String) extends Operation {
 
-  def eval(labels: Labels): Labels = op(_ + _, (s1:String, s2:String) => Some(s1 + s2), unsupportedOperation("Add"), labels)
+  def eval(labels: Labels): Labels = {
 
+    val updatedLabels: Labels = operands(labels) match {
```
cmrhmrc, line 163: updateLabels not needed, the rhs is the return value, also removes the need for the {} on eval()

app/core/models/ocelot/stanzas/CalculationStanza.scala
```diff
@@ -125,16 +158,54 @@ sealed trait Operation {
...
+
+  private def prependStringToList(l: List[String], s: String): Option[List[String]] = Some(s :: l)
+
+  private def addListToList(list1: List[String], list2: List[String]): List[String] = list1 ::: list2
 }
 
 case class SubtractOperation(left: String, right: String, label: String) extends Operation {
 
-  def eval(labels: Labels): Labels = op(_ - _, unsupportedOperation("Subtract"), subtractDate, labels)
+  def eval(labels: Labels): Labels = {
```
cmrhmrc, line 184: {} not needed on eval()

app/core/models/ocelot/stanzas/CalculationStanza.scala
```diff
@@ -125,16 +158,54 @@ sealed trait Operation {
...
+    }
+  }
 
   private def subtractDate(date: LocalDate, other: LocalDate) : Option[String] =
     Some(other.until(date, ChronoUnit.DAYS).toString)
+
+  private def subtractStringFromList(l: List[String], s: String): Option[List[String]] = Some(l.filter(_ != s))
+
+  @tailrec
+  private def subtractListFromList(list1: List[String], list2: List[String]): List[String] = {
```
cmrhmrc, line 202: {}  not required on subtractListFromList()

app/core/models/ocelot/stanzas/CalculationStanza.scala
```diff
@@ -59,25 +61,50 @@ sealed trait Operation {
...
+      case _ =>
+        val xList: Option[List[String]] = listValue(left, labels)
+        val yList: Option[List[String]] = listValue(right, labels)
+
+        (xScalar, xList, yScalar, yList)
+    }
   }
 
-  def unsupportedOperation(operationName: String)(arg1: Any, arg2: Any): Option[String] = {
+  def unsupportedOperation(operationName: String)(arg1: Any, arg2: Any ): Option[String] = {
```
cmrhmrc, line 87: these two fns could be combined and parameterized as they always return None  
`def unsupportedOperation[A](operationName: String)(arg1: Any, arg2: Any ): Option[A]`

app/core/models/ocelot/stanzas/CalculationStanza.scala
```diff
@@ -102,10 +129,16 @@ sealed trait Operation {
...
+  f(list, scalar) match {case Some(value) => labels.updateList(label, value) case None => labels}
+
+  def listOp(list1: List[String], list2: List[String], f:(List[String], List[String]) => List[String], labels: Labels): Labels =
+    labels.updateList(label, f(list1, list2))
+
   def rounding(f: (BigDecimal, Int) => BigDecimal, labels: Labels): Labels = {
 
-    val x: String = value(left, labels)
-    val y: String = value(right, labels)
+    val x: String = value(left, labels).getOrElse("")
```
cmrhmrc, line 140: Rather than getOrElse("") etc, if either of the values are missing return immediately

app/core/models/ocelot/stanzas/CalculationStanza.scala
woollett-r: Done. Updatedlabels was probably a hangover from running the debugger on the tests.

app/core/models/ocelot/stanzas/CalculationStanza.scala
woollett-r: Done

PR #260, opened 2021-02-23
EG-1296 Implement Sequence UI component and render view

woollett-r:
Note, I've used the SubmittedTextAnswerFormProvider for the time being; this may change. I have also used the identifier name[index] for the check box identifiers. I think this will work with the "list" type binding, but I'll find out more later.

### DIFF
--- a/app/forms/FormsHelper.scala
+++ b/app/forms/FormsHelper.scala
@@ -19,7 +19,7 @@ package forms
 import play.api.mvc._
 import play.api.data.{Form, Mapping}
 import play.api.data.Forms.nonEmptyText
-import core.models.ocelot.stanzas.{Input, DateInput, DataInput, Question}
+import core.models.ocelot.stanzas.{Input, DateInput, DataInput, Question, Sequence}
 import models.ui.DateInput.partitionSubmittedDateAnswer
 import models.ui.{SubmittedAnswer, SubmittedDateAnswer, SubmittedTextAnswer}
 
@@ -36,7 +36,7 @@ object FormsHelper {
   def populatedForm(inputStanza: DataInput, path: String, answer: Option[String]): Form[_] =
     inputStanza match {
       case _: DateInput => populateSubmittedDateAnswerForm(answer)
-      case _: Input | _: Question => populateSubmittedTextAnswerForm(path -> nonEmptyText, path, answer)
+      case _: Input | _: Question | _: Sequence => populateSubmittedTextAnswerForm(path -> nonEmptyText, path, answer) // TODO - Temporary solution whilst rendering sequence
     }
 
   private def bindSubmittedTextAnswer(bindData: (String, Mapping[String]))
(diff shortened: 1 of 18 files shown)


app/forms/FormsHelper.scala
```diff
@@ -36,7 +36,7 @@ object FormsHelper {
   def populatedForm(inputStanza: DataInput, path: String, answer: Option[String]): Form[_] =
     inputStanza match {
       case _: DateInput => populateSubmittedDateAnswerForm(answer)
-      case _: Input | _: Question => populateSubmittedTextAnswerForm(path -> nonEmptyText, path, answer)
+      case _: Input | _: Question | _: Sequence => populateSubmittedTextAnswerForm(path -> nonEmptyText, path, answer) // TODO - Temporary solution whilst rendering sequence
```
cmrhmrc, line 39: think this will also need adding to bindFormData above

app/services/UIBuilder.scala
```diff
@@ -196,4 +197,14 @@ class UIBuilder {
   private def fromSectionAndNoteCallout(caption: Text, nc: NoteCallout)(implicit stanzaIdToUrlMap: Map[String, String], lang: Lang): UIComponent =
     Details(caption, Seq(TextBuilder.fromPhrase(nc.text)))
 
+  private def fromSequence(sequence: Sequence, components: Seq[UIComponent])
+                          (implicit stanzaIdToUrlMap: Map[String, String], lang: Lang): UIComponent = {
+    val (errMsgs, uiElements) = partitionComponents(components, Seq.empty, Seq.empty)
+
+    val (text, hint) = TextBuilder.singleTextWithOptionalHint(sequence.text)
```
cmrhmrc, line 204: There is a bug in Question (in the backlog) to replace use of singleTextWithOptionalHint with something like fromPhrase which also extracts the hint. Inputs dont have this bug, it prevents what becomes the H1 from including label references (this is one of Abby's discoveries)

app/services/UIBuilder.scala
```diff
@@ -196,4 +197,14 @@ class UIBuilder {
   private def fromSectionAndNoteCallout(caption: Text, nc: NoteCallout)(implicit stanzaIdToUrlMap: Map[String, String], lang: Lang): UIComponent =
     Details(caption, Seq(TextBuilder.fromPhrase(nc.text)))
 
+  private def fromSequence(sequence: Sequence, components: Seq[UIComponent])
```
cmrhmrc, line 200: Looking at GDS, it supports hints on each checkbox item, so we'll need to support the [hint:....] label at the end of each checkbox item text as the main text

app/forms/FormsHelper.scala
woollett-r: This will be done as part of the submission task.

app/services/UIBuilder.scala
woollett-r: I have raised separate task for this (EG-1309)

app/services/UIBuilder.scala
woollett-r: I've updated task EG-1307 which records this defect such that sequence should also be fixed.

PR #266, opened 2021-03-11
EG-1307, added placeholder support and single hint to Question, Answer, Input and Sequence text

cmrhmrc:


app/services/TextBuilder.scala
```diff
@@ -72,13 +72,15 @@ object TextBuilder {
     (txts.head.trim, hint)
   }
 
-  // Parses a string potentially containing a hint pattern[hint:<Text Hint>]
+  // Parses current lang value of Phrase potentially containing a hint pattern[hint:<Text Hint>]
   // The string before the first hint will be converted to a Text object
   // and returned along with optional hint
   // All characters after the optional hint pattern are discarded
-  def singleTextWithOptionalHint(txt: Phrase)(implicit lang: Lang): (Text, Option[Text]) = {
+  def fromPhraseWithOptionalHint(txt: Phrase)(implicit urlMap: Map[String, String], lang: Lang): (Text, Option[Text]) = {
```
woollett-r, line 79: Oddly enough when I started looking at this problem I wrote the method "fromPhraseWithTrailingOptionalHint" which is almost identical to this.

PR #269, opened 2021-03-16
EG-1318, Replace use of page id for ValueError redirection, Maintain Sequence flow labels when using backlink

cmrhmrc:


app/core/models/ocelot/LabelCache.scala
```diff
@@ -20,8 +20,9 @@ import core.models.ocelot.stanzas.Stanza
 import play.api.i18n.Lang
 
 trait Flows {
-  def pushFlows(flowNext: Seq[String], continue: String, labelName: Option[String], labelValues: Seq[String], stanzas: Map[String, Stanza]): Labels
-  def takeFlow: Option[(String, Labels)]
+  def pushFlows(flowNext: Seq[String], continue: String, labelName: Option[String], labelValues: Seq[String], stanzas: Map[String, Stanza]): (Option[String], Labels)
```
woollett-r, line 23: Looking at SequenceStanza I would say that flowNext is a list of strings.

cmrhmrc: Yes, updated

app/core/models/ocelot/LabelCache.scala
```diff
@@ -85,24 +86,33 @@ private class LabelCacheImpl(labels: Map[String, Label] = Map(),
...
-  def takeFlow: Option[(String, Labels)] = // Remove head of flow stack and update flow label if required
-    stack.headOption.map{
-      case f: Flow =>
-      (f.next,
-       f.labelValue.flatMap(lv => lv.value.map(v => new LabelCacheImpl(labels, updateOrAddScalarLabel(lv.name, v, None), stack.tail, pool, poolCache)))
-        .getOrElse(new LabelCacheImpl(labels, cache, stack.tail, pool, poolCache))
-      )
-      case c: Continuation => (c.next, new LabelCacheImpl(labels, cache, stack.tail, pool, poolCache))
+  def nextFlow: Option[(String, Labels)] = // Remove head of flow stack and update flow label if required
+    stack match {
```
woollett-r, line 102: As I don't think x is used anywhere we could replace it with _. However, that might make things even more confusing.

cmrhmrc: I've added this update to current ticket which affects the same area

app/repositories/SessionRepository.scala
```diff
@@ -135,12 +139,12 @@ class DefaultSessionRepository @Inject() (config: AppConfig, component: Reactive
       Left(DatabaseError)
     }
 
-  def get(key: String, pageHistoryUrl: Option[String], previousPageByLink: Boolean): Future[RequestOutcome[ProcessContext]] =
+  def get(key: String, pageUrl: Option[String], previousPageByLink: Boolean): Future[RequestOutcome[ProcessContext]] =
```
woollett-r, line 142: If you are changing the name here it might be worth changing the name in the calling method in the guidance service also.

cmrhmrc: done

app/repositories/SessionProcessFSM.scala
```diff
@@ -0,0 +1,70 @@
...
+  //          no page history or flowstack updates given the update described has already taken place.
+  // forceForward, true indicates that a url similar to head of prior history (looks like a BACK) should be treated as a forward movement
+
+  // New state == (priorSp + new Pagehistory head) ++ pageHistory and flowStack output updates
+  //
+  // Output
+  // optional backlink to be displayed on page with incoming url
+  // optional page history update
+  // optional flowStack update
+  def apply(url: String, priorSp: SessionProcess, forceForward: Boolean):
```
woollett-r, line 41: IntelliJ isn't very appreciative of using the method name "apply" here. For some reason it doesn't recognise the method defined in SessionProcessFSM in the class SessionRepository.

cmrhmrc: Its not the apply, its the missing () after the @Inject() it doesnt like, I've added it and we now have our olf favourite on match statements, cyclomatic complexity

test/core/services/PageBuilderSpec.scala
```diff
@@ -24,7 +24,7 @@ import core.models.ocelot._
 import play.api.libs.json._
 import core.models.StanzaHelper
 
-class PageBuilderSpec extends BaseSpec with ProcessJson with StanzaHelper {
+class PageBuilderSpec extends BaseSpec with ProcessJson with SequenceJson with StanzaHelper {
```
woollett-r, line 27: Does PageBuilderSpec use anything from SequenceJson?

cmrhmrc: No, removed

PR #270, opened 2021-03-18
EG-1342, reset session and return first page using header link

cmrhmrc:


cmrhmrc:
Requires Journey test updates before merge

app/services/GuidanceService.scala
```diff
@@ -45,6 +45,23 @@ class GuidanceService @Inject() (
 
   val logger: Logger = Logger(getClass)
 
+  def sessionRestart(processCode: String, sessionId: String)(implicit context: ExecutionContext): Future[RequestOutcome[String]] =
+    sessionRepository.getResetSession(sessionId).map{
+      case Right(ctx) if processCode == ctx.process.meta.processCode =>
+        (ctx.urlToPageId.toList.map(e => (e._2, e._1)).toMap)
```
woollett-r, line 51: You could use map.find(_._2 = ctx.process.startPageId) which would give you Some((url,id)). Perhaps simpler.

test/core/services/PageBuilderSpec.scala
```diff
@@ -432,6 +432,34 @@ class PageBuilderSpec extends BaseSpec with ProcessJson with StanzaHelper {
...
+          Phrase(Vector("Some Text2", "Welsh, Some Text2")),
+          Phrase(Vector("Some Text3", "Welsh, Some Text3"))
+        ),
+        Vector[Link]()
+      )
+
+      pageBuilder.pagesWithValidation(process) match {
+        case Left(List(UseOfReservedUrl("4"), UseOfReservedUrl("5"))) => succeed
+        case Left(err) => fail(s"UseOfReservedUrl error not detected, failed with $err")
+        case res => fail(s"UseOfReservedUrl not detected $res")
```
woollett-r, line 459: In the unlikely event of failure it might be better to use Right(res) such that the error message is simpler.

app/services/GuidanceService.scala
cmrhmrc: done

test/core/services/PageBuilderSpec.scala
cmrhmrc: done

PR #275, opened 2021-03-30
EG-1344, Correctly handling linking between flow pages and non-flow pages

cmrhmrc:


test/repositories/SessionProcessFSMSpec.scala
```diff
@@ -264,29 +287,89 @@ class SessionProcessFSMSpec extends BaseSpec {
...
+                         PageHistory("/somepage", Nil),
+                         PageHistory("/another", Nil),
+                         PageHistory("/next", sp.flowStack))),
+               None,
+               Nil)
+    }
+
+  }
+
+  trait NestedFlowStackTest extends SequenceJson {
```
woollett-r, line 359: This trait isn't being used. Do we need some tests?

PR #277, opened 2021-03-31
EG-1347, ensure TypeError callout forall exclusive sequence pages

cmrhmrc:


app/core/models/ocelot/stanzas/SequenceStanza.scala
```diff
@@ -106,4 +106,10 @@ case class ExclusiveSequence(text: Phrase,
                              stack: Boolean) extends Sequence {
   lazy val (exclusiveOptions: Seq[Phrase], nonExclusiveOptions: Seq[Phrase]) =
     options.partition{p => exclusiveOptionRegex.findFirstMatchIn(p.english).nonEmpty}
+
+  override def validInput(value: String): Option[String] =
+    asListOfInt(value).fold[Option[String]](None){l =>
+      if (l.forall(nonExclusiveOptions.indices.contains(_)) ||
+         (l.length == 1 && l.head == nonExclusiveOptions.length)) Some(value) else None
```
woollett-r, line 113: It might be clearer to use l.head == exclusiveOptions.head.

app/core/services/PageBuilder.scala
```diff
@@ -207,20 +208,30 @@ class PageBuilder @Inject() (val placeholders: Placeholders) extends ProcessPopu
     }
 
   @tailrec
-  private def checkExclusiveSequencePages(pages: Seq[Page], errors: List[GuidanceError]): List[GuidanceError] = {
+  private def checkExclusiveSequenceTypeError(pages: Seq[Page], errors: List[GuidanceError]): List[GuidanceError] =
+    pages match {
+      case Nil => errors
+      case p :: xs if p.stanzas.collectFirst{case _: ExclusiveSequence => Unit}.fold(false)(_ => true) &&
```
woollett-r, line 214: Is it better to use () rather than Unit? I tried a similar construct in REPL and it complained about the use of Unit in source code.

app/core/models/ocelot/stanzas/SequenceStanza.scala
cmrhmrc: l.head is an index into the options seq, exclusiveOptions.head is the Phrase of the representing the exclusive option text

app/core/services/PageBuilder.scala
cmrhmrc: updated

PR #286, opened 2021-04-15
EG-1366, [list:<label>:length] placholder support

cmrhmrc:


app/core/models/ocelot/package.scala
```diff
@@ -47,6 +52,13 @@ package object ocelot {
   def pageLinkIds(phrases: Seq[Phrase]): List[String] = phrases.flatMap(phrase => pageLinkIds(phrase.english)).toList
   def labelReferences(str: String): List[String] = plSingleGroupCaptures(labelRefRegex, str)
   def labelReference(str: String): Option[String] = plSingleGroupCaptures(labelRefRegex, str).headOption
+  def listLength(listName: String, labels: Labels): Option[String] = labels.valueAsList(listName).fold[Option[String]](None){l => Some(l.length.toString)}
```
woollett-r, line 55: Should this method really be defined in Labels/LabelCacheImpl?

app/core/models/ocelot/package.scala
```diff
@@ -47,6 +52,13 @@ package object ocelot {
...
+  def labelScalarMatch(m: Regex.Match, labels: Labels, lbl: String => Option[String]): Option[String] =
+    Option(m.group(1)).fold[Option[String]]{
+      Option(m.group(4)).fold[Option[String]](None)(list => listLength(list, labels))
+    }{label => lbl(label)}
+  def labelScalarValue(str: String)(implicit labels: Labels): Option[String] =
```
woollett-r, line 60: Again, I wonder if this method should be in in Labels/LabelCacheImpl. It is the class with the primary responsibility for storing labels and giving access to their values.

app/services/UIBuilder.scala
```diff
@@ -45,23 +47,21 @@ case object ValueTypeError extends ErrorStrategy {
...
       List(expandLabelReferences(Nil), BulletPointBuilder.groupBulletPointInstructions(Nil), Aggregator.aggregateStanzas(Nil), stackStanzas(Nil))
     Page(url, fromStanzas(stanzaTransformPipeline.foldLeft(stanzas){case (s, t) => t(s)}, Nil, errStrategy.default(stanzas)))
   }
 
-  private def expandLabelReferences(acc: List[VisualStanza])(stanzas: Seq[VisualStanza])(implicit labels: Labels, lang: Lang): Seq[VisualStanza] =
+  private def expandLabelReferences(acc: List[VisualStanza])(stanzas: Seq[VisualStanza])(implicit ctx: UIBuildData): Seq[VisualStanza] =
     stanzas match {
       case Nil => acc.reverse
-      case (i: Instruction) +: xs => expandLabelReferences(i.copy(text = TextBuilder.expandLabels(i.text, labels)) :: acc)(xs)
+      case (i: Instruction) +: xs => expandLabelReferences(i.copy(text = TextBuilder.expandLabels(i.text, ctx.labels)) :: acc)(xs)
```
woollett-r, line 59: I think we need to do this for NoteCallouts as well, because like instructions they can be used to create bullet point lists. I have implemented this is in EG-1373.

app/core/models/ocelot/package.scala
cmrhmrc: think the Labels interface has all the essential fns using a "minimal and complete" criteria, this is an extra convenience fn used to implement the [list:<label>:length] placeholder when applying the regex.

app/core/models/ocelot/package.scala
cmrhmrc: as above but more so, this fn is primarily applying a regex to a string to detect use of a placeholder unrelated to the essential Labels interface

test/core/services/PageBuilderSpec.scala
```diff
@@ -68,7 +68,7 @@ class PageBuilderSpec extends BaseSpec with ProcessJson with StanzaHelper {
 
     private val phrases = Vector[Phrase](
       Phrase(Vector("Some Text", "Welsh: Some Text")),
-      Phrase(Vector(s"Some Text1 [link:Link to stanza 17:$pageId7]", s"Welsh: Some Text1 [link:Link to stanza 17:$pageId7]")),
+      Phrase(Vector(s"Some Text1 [link:Link to stanza 17:$pageId7]", s"Welsh, Some Text1 [link:Link to stanza 17:$pageId7]")),
```
woollett-r, line 71: I doesn't really matter as I don't think there are any bullet point list matches in the text, but why "Welsh,"?

app/services/UIBuilder.scala
cmrhmrc: yes , but its a forward reference to future code ;)

test/core/services/PageBuilderSpec.scala
cmrhmrc: Ha, it was intential to exercise the Fake welsh fix code in a test. The added benefit was also it increased the coverage

app/core/models/ocelot/package.scala
woollett-r: I'll accept that argument.

app/core/models/ocelot/package.scala
woollett-r: As above, but it is probably more relevant for this method.

PR #287, opened 2021-04-16
EG-1377, redirect to start of indicated process when session missing (after bookmarked page access)

cmrhmrc:


test/controllers/GuidanceControllerSpec.scala
```diff
@@ -1153,13 +1153,9 @@ class GuidanceControllerSpec extends BaseSpec with ViewFns with GuiceOneAppPerSu
...
 
-    "return a bad request response" in new Test {
+    "return a reidrect response" in new Test {
 
-      status(result) shouldBe Status.BAD_REQUEST
-    }
-
-    "be a HTML response" in new Test {
-      contentType(result) shouldBe Some("text/html")
+      status(result) shouldBe Status.SEE_OTHER
```
woollett-r, line 1158: Perhaps you could check the redirect location here. This would confirm expectation failure.

cmrhmrc: added

PR #291, opened 2021-04-21
ContainsFix

cmrhmrc:


app/core/models/ocelot/stanzas/ChoiceStanza.scala
```diff
@@ -113,7 +113,20 @@ case class ContainsTest(left: String, right: String) extends ChoiceTest {
...
+  def eval(labels: Labels): Boolean =
+    labels.valueAsList(labelReference(left).getOrElse(left)).fold(op(contains(_, _), contains(_, _), contains(_, _), labels)){l =>
+      l.exists{el =>
+        val y = value(right, labels)
+        (asDate(el), asDate(y)) match {
+          case (Some(ld1), Some(ld2)) => contains(ld1, ld2)
+          case _ =>
+            (asDecimal(el), asDecimal(y)) match {
+              case (Some(bd1), Some(bd2)) => contains(bd1, bd2)
+              case _ => contains(el.toString, y.toString)
```
woollett-r, line 125: According to REPL the list

val test: List[String] = List("One", "Two", "Three")

gives

List(One, Two, Three)

when "toString" is invoked. Using mkString(", ") would remove "List(" and ")".

app/core/models/ocelot/stanzas/ChoiceStanza.scala
```diff
@@ -113,7 +113,22 @@ case class ContainsTest(left: String, right: String) extends ChoiceTest {
...
+    labelReference(left).fold(op(contains(_, _), contains(_, _), contains(_, _), labels)){lbl =>
+      labels.valueAsList(lbl).fold(op(contains(_, _), contains(_, _), contains(_, _), labels)){l =>
+        l.exists{el =>
+          val y = value(right, labels)
+          (asDate(el), asDate(y)) match {
+            case (Some(ld1), Some(ld2)) => contains(ld1, ld2)
+            case _ =>
+              (asDecimal(el), asDecimal(y)) match {
+                case (Some(bd1), Some(bd2)) => contains(bd1, bd2)
+                case _ => contains(el.toString, y.toString)
```
woollett-r, line 126: It might be better to use mkString(", ")

cmrhmrc: el is an element of the list

woollett-r: Sorry, I mistakenly thought exists was working on an Option rather than the list. So it looks ok. However, it might be worth moving the definition of y outside the exists loop.

cmrhmrc: done

app/core/models/ocelot/stanzas/ChoiceStanza.scala
```diff
@@ -113,7 +113,22 @@ case class ContainsTest(left: String, right: String) extends ChoiceTest {
   def contains(l: BigDecimal, r: BigDecimal): Boolean = l.toString.toLowerCase().contains(r.toString.toLowerCase())
```
woollett-r, line 113: I wonder do you really need toLowerCase for BigDecimal or LocalDate

cmrhmrc: possibly not, I noticed that it was something Ocelot was doing before doing the contains test, which makes sense for strings but not decimals and dates with numeric months neither

PR #293, opened 2021-04-22
EG-1385 Exclusive sequence component accessibility fix part the first

woollett-r:


app/core/models/ocelot/package.scala
```diff
@@ -42,7 +42,8 @@ package object ocelot {
   val listOfintegerRegex: Regex = s"$commaSeparatedIntsPattern".r
   val anyIntegerRegex: Regex = "^[\\-]?\\d+$".r
   val EmbeddedParameterRegex: Regex = """\{(\d)\}""".r
-  val exclusiveOptionRegex: Regex = """\[exclusive\]""".r
+  val exclusiveOptionRegex: Regex = "\\[exclusive:([A-za-z0-9:,\\s\\-_]+)\\]".r
```
cmrhmrc, line 45: Duplicate "z", need an A-Za-z here

app/models/ui/Sequence.scala
```diff
@@ -27,6 +27,7 @@ case class ExclusiveSequence(text: Text,
                              hint: Option[Text],
                              options: Seq[Text],
                              exclusiveOption: Text,
+                             exclusiveOptionHint: Option[Text],
```
cmrhmrc, line 30: Does accessibility allow this to be optional??

app/services/UIBuilder.scala
```diff
@@ -244,12 +244,23 @@ class UIBuilder {
     val (text, hint) = TextBuilder.fromPhraseWithOptionalHint(exclusiveSequence.text)
     val options: Seq[Text] = exclusiveSequence.nonExclusiveOptions.map{phrase => TextBuilder.fromPhrase(phrase)}
 
+    val exclusiveOptionHintEnglish: Option[String] =
+      exclusiveOptionHintRegex.findAllMatchIn(exclusiveSequence.exclusiveOptions.head.english).toList.headOption.map(m => m.group(1))
```
cmrhmrc, line 248: can use findFirstMatchIn() as only one allowed

app/services/UIBuilder.scala
```diff
@@ -244,12 +244,23 @@ class UIBuilder {
...
+
+    val exclusiveOptionHintWelsh: Option[String] =
+      exclusiveOptionHintRegex.findAllMatchIn(exclusiveSequence.exclusiveOptions.head.welsh).toList.headOption.map(m => m.group(1))
```
cmrhmrc, line 251: as above

app/core/models/ocelot/package.scala
```diff
@@ -42,7 +42,8 @@ package object ocelot {
...
+  val exclusiveOptionHintRegex: Regex = "\\[exclusive:([^\\]]+)\\]".r
```
cmrhmrc, line 46: think you only need a single regex here, the latter which allows any non ] char would work

app/core/models/ocelot/package.scala
woollett-r: Regex no longer used

app/core/models/ocelot/package.scala
woollett-r: Done

app/models/ui/Sequence.scala
woollett-r: The optionality arises from the use of regex matching. In practice, owing to the mechanism used to identify exclusive sequences the hint will always be defined. I have kept the optionality as it provides a useful way of distinguishing between exclusive and non-exclusive sequences when rendering the sequence checkboxes.

app/services/UIBuilder.scala
woollett-r: I've implemented findFirstMatchIn.

app/services/UIBuilder.scala
woollett-r: As above

PR #296, opened 2021-04-27
EG-1395, added error name at start of error message

cmrhmrc:


cmrhmrc:
Will need journey tests update

PR #302, opened 2021-05-05
EG-1405, map wrong session found to expectation failed error => resta…

cmrhmrc:
…rt process

### DIFF
--- a/app/services/GuidanceService.scala
+++ b/app/services/GuidanceService.scala
@@ -112,7 +112,7 @@ class GuidanceService @Inject() (
           )
         }
       case Right(_) =>
-        logger.error(s"Referenced session ( $sessionId ) does not contain a process with processCode $processCode")
+        logger.warn(s"Referenced session ( $sessionId ) does not contain a process with processCode $processCode")
         Left(ExpectationFailedError)
       case Left(err) =>
         logger.error(s"Repository returned $err, when attempting retrieve process using id (sessionId) $sessionId")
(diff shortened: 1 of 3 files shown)


app/services/GuidanceService.scala
```diff
@@ -112,7 +113,7 @@ class GuidanceService @Inject() (
         }
       case Right(_) =>
         logger.error(s"Referenced session ( $sessionId ) does not contain a process with processCode $processCode")
```
woollett-r, line 115: I think there is a case for reducing the logging level to warn. We are not handling an internal error rather it is a case of unsupported functionality.

test/controllers/GuidanceControllerSpec.scala
```diff
@@ -1152,15 +1152,15 @@ class GuidanceControllerSpec extends BaseSpec with ViewFns with GuiceOneAppPerSu
           mockGuidanceService,
           stubMessagesControllerComponents()
         )
-      lazy val result = target.getPage(processCode, path, None)(fakeRequest)
+      lazy val result = target.getPage("otherProcessCode", "/path", None)(fakeRequest)
     }
 
     "return a reidrect response" in new Test {
```
woollett-r, line 1158: Spelling

test/controllers/GuidanceControllerSpec.scala
```diff
@@ -1133,7 +1133,7 @@ class GuidanceControllerSpec extends BaseSpec with ViewFns with GuiceOneAppPerSu
```
woollett-r, line 1133: Strictly speaking the update is in  GuidanceService so there should be a new or updated test in GuidanceServiceSpec.

PR #305, opened 2021-05-14
EG-1430 Implement date output translation

woollett-r:


app/services/TextBuilder.scala
```diff
@@ -21,7 +21,7 @@ import core.models.ocelot.{Link, Phrase, hintRegex, labelPattern, listPattern, l
 import models.ui._
 import scala.util.matching.Regex
 import Regex._
-import play.api.i18n.Lang
+import play.api.i18n.{Lang, MessagesApi}
 import scala.annotation.tailrec
 
 object TextBuilder {
```
cmrhmrc, line 27: if you inject Messages into TextBuilder rather than UIBuilder, the TextBuilder fns signatures would stay as they are and the scope of the messages would be limited to only TextBuilder (appropriately) and OutpuFormat

PR #306, opened 2021-05-14
EG-1429, valid all incoming pages are valid for current context

cmrhmrc:


app/controllers/GuidanceController.scala
```diff
@@ -93,12 +93,12 @@ class GuidanceController @Inject() (
       case Left(NotFoundError) =>
         logger.warn(s"Request for PageContext at /$path returned NotFound, returning NotFound")
         Future.successful(NotFound(errorHandler.notFoundTemplateWithProcessCode(Some(processCode))))
-      case Left(BadRequestError) =>
-        logger.warn(s"Request for PageContext at /$path returned BadRequest")
-        Future.successful(BadRequest(errorHandler.badRequestTemplateWithProcessCode(Some(processCode))))
       case Left(ExpectationFailedError) =>
         logger.error(s"Redirecting to start of processCode $processCode at ${appConfig.baseUrl}/$processCode")
         Future.successful(Redirect(s"${appConfig.baseUrl}/$processCode"))
+      case Left(ForbiddenError) =>
```
woollett-r, line 99: probably worth updating logger statements such the we know the type of error that occurred. Expectation failed and forbidden are using the same message.

app/repositories/SessionRepository.scala
```diff
@@ -130,57 +128,109 @@ class DefaultSessionRepository @Inject() (config: AppConfig,
...
+            logger.warn(s"Attempt to move to unknown page $url")
+            Future.successful(Left(NotFoundError))
+          }{pageDesc =>
+            logger.debug(s"Incoming Page: ${pageDesc.id}, $url, current legalPageIds: ${sp.legalPageIds}")
+            if (sp.legalPageIds.isEmpty || sp.legalPageIds.contains(pageDesc.id)){ // Wild card or fixed list of valid page ids
+              val firstPageUrl: String = s"${sp.process.meta.processCode}${sp.process.startUrl.getOrElse("")}"
+              val (backLink, historyUpdate, flowStackUpdate, labelUpdates) = sessionProcessTransition(url, sp, previousPageByLink, firstPageUrl)
+              val labels: Map[String, Label] = sp.labels ++ labelUpdates.map(l => (l.name -> l)).toMap
+              val legalPageIds = Process.StartStanzaId :: pageDesc.id :: pageDesc.next ++
+                                  backLink.fold[List[String]](Nil)(bl => List(sp.pageMap(bl.drop(sp.process.meta.processCode.length)).id)).distinct
```
woollett-r, line 205: I suspect "distinct" is only being applied to "(bl => List(sp.pageMap(bl.drop(sp.process.meta.processCode.length)).id))". You need to put some brackets around the whole statement and then apply distinct.

app/repositories/SessionRepository.scala
```diff
@@ -130,57 +128,109 @@ class DefaultSessionRepository @Inject() (config: AppConfig,
...
+                                    sp.continuationPool,
+                                    sp.pageMap,
+                                    legalPageIds,
+                                    backLink)
+              logger.debug(s"NewLegalIds: ${legalPageIds}")
+              saveUpdates(key, historyUpdate, flowStackUpdate, labelUpdates, legalPageIds).map {
                 case Left(err) =>
                   logger.error(s"Unable to save backlink history, error = $err")
                   Right(processContext)
                 case _ => Right(processContext)
```
woollett-r, line 220: Should we return the processContext here or an error. If something goes wrong with a flow update or label updates should we continue?

app/repositories/SessionRepository.scala
```diff
@@ -130,57 +128,109 @@ class DefaultSessionRepository @Inject() (config: AppConfig,
...
-          Left(DatabaseError)
-      }
+    }.recover {
+      case lastError =>
+        logger.error(s"Error $lastError while trying to retrieve process from session repo with _id=$key")
+        Left(DatabaseError)
+    }
 
   def getResetSession(key: String): Future[RequestOutcome[ProcessContext]] =
     findAndUpdate(
```
woollett-r, line 236: When updating should we reset the legal page ids as well?

app/services/GuidanceService.scala
```diff
@@ -183,18 +187,25 @@ class GuidanceService @Inject() (
...
           logger.warn(s"Failed to parse process with error $err")
           Future.successful(Left(InvalidProcessError))
         },
-        pages => sessionRepository.set(docId, process, pages.map(p => p.url -> p.id).toMap).map {
-          case Right(_) => Right((pages.head.url, process.meta.processCode))
-          case Left(err) =>
-            logger.error(s"Failed to store new parsed process in session repository, $err")
-            Left(err)
+        pages => {
+          val urlMap: Map[String, String] = pages.map(p => (p.id, p.url)).toMap
```
woollett-r, line 195: As you are doing a bit of work here it would be a good idea to put the debug logging inside an if block as follows

if(logger.isDebugEnabled()) {
.....
.....
}
Then when you aren't debugging the code won't get executed

app/controllers/GuidanceController.scala
cmrhmrc: done

app/repositories/SessionRepository.scala
cmrhmrc: done

app/services/GuidanceService.scala
cmrhmrc: done

app/repositories/SessionRepository.scala
cmrhmrc: done

test/services/GuidanceServiceSpec.scala
```diff
@@ -411,12 +410,31 @@ class GuidanceServiceSpec extends BaseSpec  with GuiceOneAppPerSuite {
       }
 
     }
+  }
+
+  "Calling getProcessContext full" should {
+
+    // "return a Process context containing legalPageIds" in new Test {
```
woollett-r, line 417: Should this test be removed or corrected?

cmrhmrc: beginnings of further tests

PR #316, opened 2021-07-05
DL-5703 Hint text support for checkboxes

jmountain-hmrc:


app/core/models/ocelot/package.scala
```diff
@@ -44,7 +44,7 @@ package object ocelot {
   val listOfintegerRegex: Regex = s"$commaSeparatedIntsPattern".r
   val anyIntegerRegex: Regex = "^[\\-]?\\d+$".r
   val EmbeddedParameterRegex: Regex = """\{(\d)\}""".r
-  val exclusiveOptionRegex: Regex = "\\[exclusive:([^\\]]+)\\]".r
+  val exclusiveOptionRegex: Regex = "\\[exclusive]".r
```
cmrhmrc, line 47: not a valid regex ([] are special and only the first is escaped). can probably replace with a literal as its now a constant

app/models/ui/Sequence.scala
```diff
@@ -16,18 +16,10 @@
 
 package models.ui
 
-case class NonExclusiveSequence(text: Text,
-                                hint: Option[Text],
-                                options: Seq[Text],
-                                body: Seq[UIComponent],
-                                errorMsgs: Seq[ErrorMsg])
-  extends NonExclusiveSequenceFormComponent
+case class SequenceAnswer(text: Text, hint: Option[Text], exclusive: Boolean = false) extends UIComponent
```
cmrhmrc, line 19: In a set of answers, there can only ever be one excusive answer, would be useful to have the Sequence/Answer types reflect ( the boolean flag would allow multiple exclusive answers)

app/services/UIBuilder.scala
```diff
@@ -241,38 +241,21 @@ class UIBuilder {
...
-    ui.NonExclusiveSequence(text, hint, options, uiElements, errMsgs)
-  }
-
-  private def fromExclusiveSequence(exclusiveSequence: ExclusiveSequence, components: Seq[UIComponent])(implicit ctx: UIContext): UIComponent = {
-
-    val (errMsgs, uiElements) = partitionComponents(components, Seq.empty, Seq.empty)
-    val (text, hint) = TextBuilder.fromPhraseWithOptionalHint(exclusiveSequence.text)
-    val options: Seq[Text] = exclusiveSequence.nonExclusiveOptions.map{phrase => TextBuilder.fromPhrase(phrase)}
+  private def fromSequence(sequence: Sequence, components: Seq[UIComponent])(implicit ctx: UIContext): UIComponent = {
+    val answers = sequence.options.map { ans =>
```
cmrhmrc, line 245: There is validation in the backend which prevents guidance containing multiple execlusive answers (generates a MultipleExclusiveOptions back to the Ocelot user if they try to push/preview a process). So code needs only to find and process the single exclusive

app/services/UIBuilder.scala
```diff
@@ -241,38 +241,21 @@ class UIBuilder {
...
+      val isExclusive: Boolean = exclusiveOptionRegex.findFirstMatchIn(ans.english).nonEmpty
+      val exclusivePhrase = Phrase(
```
cmrhmrc, line 247: not always an exclusive phrase

app/views/components/sequence_checkbox.scala.html
```diff
@@ -28,12 +28,34 @@
...
+            ""
+        }
+
+        Html(dataExclusive + s"aria-describedby='$hintClass'")
+    } else {
+        Html(None)
+    }
+}
+
+@itemChecked(idx: Int) = @{form.data.valuesIterator.contains(idx.toString)}
```
cmrhmrc, line 53: this is generating 
<input class="govuk-checkboxes__input" id="start-3" name="start[3]" type="checkbox" value="3"  data-exclusive='true'aria-describedby='exclusive-hint'>

note no space before aria-describedby, also better to remain consistent in quoting with "" rather than ''

original line 
<input class="govuk-checkboxes__input" id="current-3" name="current[3]" type="checkbox" value="3"   data-exclusive="true" aria-describedby="exclusive-hint">

app/models/ui/Sequence.scala
```diff
@@ -16,18 +16,10 @@
...
-                             exclusiveOption: Text,
-                             exclusiveOptionHint: Option[Text],
-                             body: Seq[UIComponent],
-                             errorMsgs: Seq[ErrorMsg]
-                            ) extends ExclusiveSequenceFormComponent
+case class Sequence(text: Text,
+                    hint: Option[Text],
+                    options: Seq[SequenceAnswer],
+                    body: Seq[UIComponent],
+                    errorMsgs: Seq[ErrorMsg]) extends SequenceFormComponent
```
cmrhmrc, line 25: as discussed, we can remove the SequenceFormComponent as now redundant

app/services/UIBuilder.scala
```diff
@@ -75,8 +75,8 @@ class UIBuilder {
       case (c: Callout) :: xs => fromStanzas(xs, acc ++ fromCallout(c, errStrategy), errStrategy)
       case (in: Input) :: _ => fromStanzas(Nil, Seq(fromInput(in, acc)), errStrategy)
       case (q: Question) :: _ => fromStanzas(Nil, Seq(fromQuestion(q, acc)), errStrategy)
-      case (ns: NonExclusiveSequence) :: _ => fromStanzas(Nil, Seq(fromNonExclusiveSequence(ns, acc)), errStrategy)
-      case (es: ExclusiveSequence) :: _ => fromStanzas(Nil, Seq(fromExclusiveSequence(es, acc)), errStrategy)
+      case (ns: NonExclusiveSequence) :: _ => fromStanzas(Nil, Seq(fromSequence(ns, acc)), errStrategy)
```
cmrhmrc, line 78: as discussed, given you've simplified to a single core Sequence class, could do someting similar with the teo UI classes

app/core/models/ocelot/package.scala
jmountain-hmrc: findFirstMatchIn does expect a Regex, and it does seem to function properly as a regex too (typing anything that's an invalid regex shows up on intellij as such)

app/models/ui/Sequence.scala
jmountain-hmrc: This can be changed, but it is an additional refactor for tests and such down the line

app/services/UIBuilder.scala
jmountain-hmrc: It depends if the UI sequence model is changed then that this will be changed with it to a partition (using the first of the exclusive options)

app/views/components/sequence_checkbox.scala.html
jmountain-hmrc: Tricky to fix the quotations but I can try.

app/core/models/ocelot/package.scala
cmrhmrc: you've got away with it, if you want to use [ or ] literally rather than their meaning in a regex, they should be escaped, so it works, but its misleading. What I mean by replace with a literal, is using the features of regular expressions, findFirstMatchIn etc. is no longer needed as its been reduced to simple text match and replace i.e could use contains and replaceAll, replaceAllLiterally on String/StringOps. If you stick with regex, there should be a \\ before the second ]

app/models/ui/Sequence.scala
cmrhmrc: it would improve the code, I can help with test changes

app/services/UIBuilder.scala
cmrhmrc: dont understand your answer???

app/views/components/sequence_checkbox.scala.html
cmrhmrc: you can use view.components.ariaDescribedBy.scala.html I think

app/core/models/ocelot/stanzas/SequenceStanza.scala
```diff
@@ -52,7 +52,7 @@ object SequenceStanza {
     )(unlift(SequenceStanza.unapply))
 }
 
-trait Sequence extends VisualStanza with Populated with DataInput {
+trait SequenceLike extends VisualStanza with Populated with DataInput {
```
cmrhmrc, line 55: SequenceLike no longer need, it content can be moved into the Sequence case class

cmrhmrc (changes requested):
Just the merging of Sequence and SequenceLike

PR #339, opened 2021-11-09
DL-6595, upgrade to Play 2,8

cmrhmrc:


cmrhmrc:
scalatest dependency left at 3.1.1 to avoid dependency issue with com/vladsch/flexmark/util/ast/Node

PR #343, opened 2021-11-26
DL-6440, Session updates moved into GuidanceService and Init fns moved into RetrieveAndCache service

cmrhmrc:


app/services/SessionService.scala
```diff
@@ -0,0 +1,54 @@
...
+// import repositories.SessionRepository
+// import core.models.ocelot.{LabelCache, Labels, Process}
+
+// @Singleton
+// class SessionService @Inject() (
+//     appConfig: AppConfig,
+//     sessionRepository: SessionRepository,
+//     messagesApi: MessagesApi
+// ) {
+
```
paulanderson1234: Any value in this file if everything is commented out?

cmrhmrc: This file is deleted

test/models/ProcessContextSpec.scala
```diff
@@ -23,7 +23,7 @@ import play.api.inject.Injector
 import org.scalatestplus.play.guice.GuiceOneAppPerSuite
 import services.SecuredProcessBuilder
 
-class ProcessContextSpec extends BaseSpec with GuiceOneAppPerSuite {
+class GuidanceSessionSpec extends BaseSpec with GuiceOneAppPerSuite {
```
paulanderson1234, line 26: Would we not rename this file instead of leaving it as ProcessContextSpec' ?

cmrhmrc: renamed and pushed to branch

app/services/SessionService.scala
paulanderson1234: Confirmed - resolved

test/models/ProcessContextSpec.scala
paulanderson1234: Resolved

PR #354, opened 2022-02-07
DL-7178 Add Optimizely code snippet

MarjanMubarok:


app/views/main_layout.scala.html
```diff
@@ -72,6 +72,7 @@
     <!--[if lte IE 8]><link href=' @controllers.routes.Assets.versioned("stylesheets/application-ie-8.css")' rel="stylesheet" type="text/css" /><![endif]-->
     <!--[if gt IE 8]><!--> <link href='@controllers.routes.Assets.versioned("stylesheets/application.css")' media="all" rel="stylesheet" type="text/css" />
         <!--<![endif]-->
+    <script src="https://cdn.optimizely.com/js/21016452515.js"></script>
```
cmrhmrc, line 75: can make this conditional on the presence of the optimizely id

app/views/main_layout.scala.html
```diff
@@ -72,7 +72,9 @@
...
```
cmrhmrc, line 75: this line should now be deleted as the script is now included conditionally below

PR #379, opened 2022-08-05
DL-8345 Add view test for guidance error Page

MarjanMubarok:


test/views/GuidanceErrorSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+      val errorParagraph = doc.getElementsByTag("p").text
+      val solutionParagraph = doc.getElementsByTag("p").text
+
+      errorParagraph contains "NonTerminatingPageError: Guidance contains non-terminating loop which includes stanza 4"
+      solutionParagraph contains "UnsupportedOperationError: Typically this error occurs where one or both operands are references to labels which have not been assigned a value."
+
+    }
+  }
+
+  "Display the correct error message for a single error with multiple solutions" in new Test {
```
cmrhmrc, line 59: Its multiple errors with a single soln

test/views/GuidanceErrorSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+    "Display the correct error message for a single error" in new Test {
+      val errorList: List[String] = List("NonTerminatingPageError: Guidance contains non-terminating loop which includes stanza 4")
+      val solutionsList: List[List[String]] = List(List("NonTerminatingPageError: Ensure the page terminates for all input values (if any)"))
+      val html: Html = guidanceErrorPage(pageTitle, heading, "123", errorList, solutionsList)
+      val doc = asDocument(html)
+
+      val errorParagraph = doc.getElementsByTag("p").text
+      val solutionParagraph = doc.getElementsByTag("p").text
+
+      errorParagraph contains "NonTerminatingPageError: Guidance contains non-terminating loop which includes stanza 4"
```
cmrhmrc, line 53: better to use the messages api here otherwise changes to the message files will mean fixes for the tests

test/views/GuidanceErrorSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+    "1. Check your answers page. Three columns, column 1 cells must not be bold and every column 3 is a link.",
+    "2. Summary list. Two columns and column 1 cells must not be bold",
+    "3. Table. At least one column and two rows. The cells of row 1 are headings, all of which must be bold. The Row stanzas which make up the table must be stacked to a Callout of type SubSection which will become the overall title of the table"))
+    val html: Html = guidanceErrorPage(pageTitle, heading, "123", errorList, solutionsList)
+    val doc = asDocument(html)
+
+    val errorParagraph = doc.getElementsByTag("p").text
+    val solutionParagraph = doc.getElementsByTag("p").text
+
+    errorParagraph contains "NonTerminatingPageError: Guidance contains non-terminating loop which includes stanza 4"
```
cmrhmrc, line 72: As above, use the messages api

cmrhmrc (changes requested):
See inline comments

test/views/GuidanceErrorSpec.scala
```diff
@@ -0,0 +1,98 @@
...
+    "Display the correct error message for a single error" in new Test {
+      val errorList: List[String] = List(Messages("guidance.error.nonterminating_loop", 4))
+      val solutionsList: List[List[String]] = List(List(Messages("guidance.error.nonterminating_loop.soln")))
+      val html: Html = guidanceErrorPage(pageTitle, heading, "123", errorList, solutionsList)
+      val doc = asDocument(html)
+
+      val errorParagraph = doc.getElementsByTag("p").text
+      val solutionParagraph = doc.getElementsByTag("p").text
+
+      errorParagraph contains "NonTerminatingPageError: Infinite loop found on page containing stanza '4'"
```
cmrhmrc, line 53: contains will only return true or false, but cause a test failure, so these tests will all always pass

test/views/GuidanceErrorSpec.scala
```diff
@@ -0,0 +1,98 @@
...
+            Messages("guidance.error.unsupported_ui_pattern.soln2"),
+            Messages("guidance.error.unsupported_ui_pattern.soln3"),
+            Messages("guidance.error.unsupported_ui_pattern.soln4")))
+    val html: Html = guidanceErrorPage(pageTitle, heading, "123", errorList, solutionsList)
+    val doc = asDocument(html)
+
+    val errorParagraph = doc.getElementsByTag("p").text
+    val solutionParagraph = doc.getElementsByTag("p").text
+
+    errorParagraph contains "NonTerminatingPageError: Guidance contains non-terminating loop which includes stanza '4'"
```
cmrhmrc, line 73: as above

test/views/GuidanceErrorSpec.scala
```diff
@@ -0,0 +1,98 @@
...
+        Messages("guidance.error.unsupported_operation",4,"AddOperation","[label:X]","[label:Y]"))
+    val solutionsList: List[List[String]] = List(List(Messages("guidance.error.unsupported_operation.soln")))
+    val html: Html = guidanceErrorPage(pageTitle, heading, "123", errorList, solutionsList)
+    val doc = asDocument(html)
+
+    val errorParagraph = doc.getElementsByTag("p").text
+    val solutionParagraph = doc.getElementsByTag("p").text
+
+
+    errorParagraph contains "Calculation stanza '2' contains operation 'DivideOperation' with invalid arguments '[label:X]', '[label:Y]'."
```
cmrhmrc, line 92: as above , tests will never fail

cmrhmrc (changes requested):
Use of contains should be replaced with shouldBe otherwise tests will always pass.
Tests should reference Messages rather than hard coded text

PR #450, opened 2024-02-23
DL-12946 Fix for url appearing on print preview

StuartJamesProctor:


paulanderson1234:
This has completely removed the title and link from any printed page. We need to retain the title part as this is apparently what the business wanted

